PR #47, opened 2025-08-31
updated to include multi score metrics

ShriyaaNarayanan:
- `get_results` now returns a dictionary of scores instead of one number
- `TerminusAdapter `can aggregate multiple scores using different strategies
- Raw multi-scores are maintained in trajectories for analysis
- GEPA compatibility is maintained, however it still sees single scores, if not this would require significant architectural changes
- User has control over how scores are combined
- GEPA still gets single scores to optimize, but there is now multi-score data stored in the trajectories that can be analyzed later

### DIFF
--- a/src/gepa/adapters/terminal_bench_adapter/terminal_bench_adapter.py
+++ b/src/gepa/adapters/terminal_bench_adapter/terminal_bench_adapter.py
@@ -4,222 +4,312 @@ import subprocess
 from datetime import datetime
 from pathlib import Path
 
+
 from pydantic import BaseModel
 from terminal_bench.agents.terminus_1 import CommandBatchResponse
+
 
 from gepa import EvaluationBatch, GEPAAdapter
 
 
+
+
 class TerminalBenchTask(BaseModel):
-    task_id: str
-    model_name: str
+   task_id: str
+   model_name: str
+
+
 
 
 def run_agent_tb(
-    task_ids: str | list[str],
-    run_id: str,
-    model_name: str,
-    instruction_prompt: str,
-    dataset_name: str = "terminal-bench-core",
-    dataset_version: str = "head",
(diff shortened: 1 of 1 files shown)


LakshyAAAgrawal:
Hi @ShriyaaNarayanan ,

The goal of the issue was to make changes to the gepa codebase, such that the pareto frontier that gepa tracks has the multiple scores that the user provides. basically, the goal was not to make a change in the adapters, but to the gepa/core itself.

LakshyAAAgrawal:
Let's also revert all changes to the adapters for now

LakshyAAAgrawal:
It looks like you are solving different issues in this.

Let's make a separate PR to solve each issue.

src/gepa/utils/stopping.py
LakshyAAAgrawal, line 1: This file is great! I like this. Perhaps this should be a part of a separate PR btw.

Let's also rename it to StopCondition. So "TimeoutStopper" -> "TimeoutStopCondition"

ShriyaaNarayanan: Okay sounds good!

LakshyAAAgrawal:
@ShriyaaNarayanan feel free to prioritize the budget and stop condition PR. That can be merged immediately.

src/gepa/core/state.py
```diff
@@ -14,10 +14,12 @@ class GEPAState(Generic[RolloutOutput]):
     parent_program_for_candidate: list[list[int | None]]
 
     program_full_scores_val_set: list[float]
+    program_multi_scores_val_set: list[list[dict[str, float]]]
```
LakshyAAAgrawal, line 17: I don't think these should be separate fields. program_full_scores_val_set and prog_candidate_val_subscores should continue to be same, their types should be list[float] and list[list[tuple[float, ...]]]

PR #49, opened 2025-09-02
adding api_key parameter to AnyMaths adapter

Lucas-Fernandes-Martins:
Small change to allow non-ollama endpoints in AnyMaths Adapter. 

Tested locally and seems to be working fine.


egmaminta:
Hi @Lucas-Fernandes-Martins . Have you tried setting the API key as env variable and check if the code still works? It's one of the reasons why I didn't add api_key argument.

My latest commit (to be PR approved) also corresponds to these changes. You might have pulled my outdated commit.

Lucas-Fernandes-Martins:
Hi @egmaminta 

Yes, it works, it's just that I was using different APIs for rollouts and reflection and felt it was easier to organize by always passing api_key to LLMLite, instead of setting a default api_key for inference by the env variable and then forcing api_key onto reflection model. Also if by any chance the user wants to use an endpoint LLMLite doesn't explicitly support via env vars.
 
But if you've already added it in a future PR then awesome :) I'll pull from your fork instead

Thanks for taking a look! Will close this PR since this will be fixed either way.

PR #51, opened 2025-09-03
Introduce mlflow

### LINKED ISSUE #36: Support MLFlow

TomeHirata:
This PR introduces a remote logging approach with MLflow 

Closes #36


src/gepa/api.py
```diff
@@ -38,8 +38,11 @@ def optimize(
     logger: LoggerProtocol | None = None,
     run_dir: str | None = None,
     use_wandb: bool = False,
+    use_mlflow: bool = False,
```
LakshyAAAgrawal: Can you put all mlflow args together, so after current line 43.

Also, does mlflow not take many other args? wandb can configure many things, so I had introduced init_kwargs. I will let you decide!

uv.lock
```diff
@@ -118,6 +118,21 @@ wheels = [
     { url = "https://files.pythonhosted.org/packages/fb/76/641ae371508676492379f16e2fa48f4e2c11741bd63c48be4b12a6b09cba/aiosignal-1.4.0-py3-none-any.whl", hash = "sha256:053243f8b92b990551949e63930a839ff0cf0b0ebbe0597b0f3fb19e1a0fe82e", size = 7490, upload-time = "2025-07-03T22:54:42.156Z" },
 ]
 
+[[package]]
+name = "alembic"
```
LakshyAAAgrawal, line 122: Could you please recheck and ensure uv.lock only contains mlflow related changes?

src/gepa/logging/experiment_tracker.py
```diff
@@ -0,0 +1,146 @@
...
+        mlflow_experiment_name: Experiment name for mlflow
+        
+    Returns:
+        ExperimentTracker instance
+        
+    Raises:
+        ValueError: If both or neither backend is specified
+    """
+    if use_wandb and use_mlflow:
+        raise ValueError("Cannot use both wandb and mlflow simultaneously. Choose one.")
```
LakshyAAAgrawal: Is there a particular reason to explicitly disable this?

uv.lock
TomeHirata: They are transitive/direct dependencies of mlflow. I think the current setup is fine since mlflow was not included as a core dependency of gepa. We could remove it from the optional dependencies too if needed

src/gepa/api.py
TomeHirata: MLflow allows users to configure things in the tracking server setup. Once the mlflow tracking server is up, users just need to configure `tracking_uri` and `experiment_name`

LakshyAAAgrawal: Amazing!

src/gepa/logging/experiment_tracker.py
TomeHirata: I just felt it'd be odd to use them together, but yeah there's no harm to support both

src/gepa/api.py
```diff
@@ -196,10 +204,17 @@ def evaluator(inputs, prog):
...
+        use_mlflow=use_mlflow,
+        mlflow_tracking_uri=mlflow_tracking_uri,
+        mlflow_experiment_name=mlflow_experiment_name,
         track_best_outputs=track_best_outputs,
         display_progress_bar=display_progress_bar,
         raise_on_exception=raise_on_exception,
     )
-    state = engine.run()
+
+    # Use context manager for proper resource cleanup
```
LakshyAAAgrawal: Which resources is engine using? Also I couldn't see an __enter__, __exit__ methods in GEPAEngine.

src/gepa/logging/experiment_tracker.py
LakshyAAAgrawal, line 1: Thank you very much for this! Amazing abstraction

src/gepa/api.py
TomeHirata: Good catch, I just forgot to revert the change here. Fixed

src/gepa/proposer/reflective_mutation/reflective_mutation.py
```diff
@@ -104,6 +113,9 @@ def propose(self, state: GEPAState) -> CandidateProposal | None:
         if self.use_wandb:
             import wandb  # type: ignore
             wandb.log({"subsample_score": sum(eval_curr.scores)}, step=i)
+        elif self.use_mlflow:
```
LakshyAAAgrawal: Wouldn't it be better to use `self.experiment_tracker.log_metrics` after defining it, instead of having if/elif for wandb and mlflow, since that is already handled in experiment_tracker class?

TomeHirata: Great call, I moved the logic and propagate experiment_tracker to the downstream components.

LakshyAAAgrawal:
Do you think it might be better to extend the LoggerProtocol class (https://github.com/gepa-ai/gepa/blob/main/src/gepa/logging/logger.py#L7) with a log_metrics method, and implement that class for MLFlow, WandB, and standard logging instead? That way we can avoid sending logger and experiment_tracker as arguments in so many places, since all the places that mlflow/wandb are being used coincide with logger usage.

Or do you think it may be bad coupling.

TomeHirata:
I think that is not necessary, as `LoggerProtocol` is application logging for stdout/stderr while mlflow and wandb is for metrics logging.

TomeHirata:
Added tests for changes!

tests/test_experiment_tracking.py
```diff
@@ -0,0 +1,616 @@
+import os
+import shutil
+import sys
+import tempfile
+from pathlib import Path
+
+import pytest
+
+# Add src to path for imports
+sys.path.insert(0, "src")
```
LakshyAAAgrawal: I think this shouldn't be required?

LakshyAAAgrawal:
LGTM! Can merge, can you just look at the one comment above?

tests/test_experiment_tracking.py
TomeHirata: Removed.

PR #52, opened 2025-09-03
Add custom component selector support to optimize()

andressrg:
Allow users to pass a custom ReflectionComponentSelector to the optimize() function via the module_selector parameter. If None is provided, defaults to RoundRobinReflectionComponentSelector to maintain backward compatibility.

This addresses the feature request in stanfordnlp/dspy#8755 for customizable component selection logic in GEPA.


src/gepa/api.py
```diff
@@ -29,6 +29,8 @@ def optimize(
     skip_perfect_score=True,
     reflection_minibatch_size=3,
     perfect_score=1,
+    # Component selection configuration
+    module_selector: "ReflectionComponentSelector | None" = None,
```
LakshyAAAgrawal: Let's make this "ReflectionComponentSelector | str | None"

Such that there are a few default component selectors that GEPA uses by default.

For example, look at candidate_selection_strategy.

This allows it to be very easily specified in command line.

LakshyAAAgrawal: A much more scalable way is to have some kind of a "registry", which is basically a dict() mapping these strings to the classes. I don't think that needs to be part of this PR. There will be a much larger refactoring of the codebase to centralize all kinds of configs which is currently split across all files.

andressrg: sure!

I like the registry idea 🤔 

Started by adding a simple map between cls and str.

LakshyAAAgrawal:
LGTM! Let me know if you are changing anything else, or I should merge and version-bump

LakshyAAAgrawal:
Do you think you can add any unittests for this?

andressrg:
Sure! I can add unittests later today 👍

andressrg:
Tests added and PR rebased

Ready to go 👍

LakshyAAAgrawal (approved):

PR #65, opened 2025-09-06
Multi score

ShriyaaNarayanan:
Isolated changes from stop condition PR and only have included multi score PR changes


src/gepa/utils/stopping.py
LakshyAAAgrawal, line 1: Let's get to this PR after the stopper PR. Remove these changes from here.

Issue #66, opened 2025-09-07
ComponentSelector protocol should be a __call__


LakshyAAAgrawal:
https://github.com/gepa-ai/gepa/blob/main/src/gepa/proposer/reflective_mutation/base.py#L16 should be a __call__ instead of select_modules.

@andressrg

PR #67, opened 2025-09-07
Add 'all' component selector for selecting all components

andressrg:
Implement `all_reflection_component_selector` that returns all components for modification instead of cycling through them individually like the existing round_robin selector.

Changes:
- Add all_reflection_component_selector function in component_selector.py
- Register 'all' selector in api.py module_selector mapping
- Update docstring to document new 'all' option
- Add tests

## Related
Fixes https://github.com/gepa-ai/gepa/issues/66

### DIFF
--- a/src/gepa/api.py
+++ b/src/gepa/api.py
@@ -16,8 +16,8 @@ from gepa.proposer.reflective_mutation.reflective_mutation import ReflectiveMuta
 from gepa.strategies.batch_sampler import EpochShuffledBatchSampler
 from gepa.strategies.candidate_selector import CurrentBestCandidateSelector, ParetoCandidateSelector
 from gepa.strategies.component_selector import (
-    all_reflection_component_selector,
-    round_robin_reflection_component_selector,
+    AllReflectionComponentSelector,
+    RoundRobinReflectionComponentSelector,
 )
 
 
@@ -169,16 +169,16 @@ def optimize(
     )
 
     if isinstance(module_selector, str):
-        module_selector_fn = {
-            "round_robin": round_robin_reflection_component_selector,
-            "all": all_reflection_component_selector,
+        module_selector_cls = {
+            "round_robin": RoundRobinReflectionComponentSelector,
+            "all": AllReflectionComponentSelector,
         }.get(module_selector)
 
-        assert module_selector_fn is not None, (
+        assert module_selector_cls is not None, (
             f"Unknown module_selector strategy: {module_selector}. Supported strategies: 'round_robin', 'all'"
         )
 
-        module_selector = module_selector_fn
+        module_selector = module_selector_cls()
 
     batch_sampler = EpochShuffledBatchSampler(minibatch_size=reflection_minibatch_size, rng=rng)
 
--- a/src/gepa/api.py
+++ b/src/gepa/api.py
@@ -15,7 +15,10 @@ from gepa.proposer.reflective_mutation.base import LanguageModel, ReflectionComp
 from gepa.proposer.reflective_mutation.reflective_mutation import ReflectiveMutationProposer
 from gepa.strategies.batch_sampler import EpochShuffledBatchSampler
 from gepa.strategies.candidate_selector import CurrentBestCandidateSelector, ParetoCandidateSelector
-from gepa.strategies.component_selector import AllReflectionComponentSelector, RoundRobinReflectionComponentSelector
+from gepa.strategies.component_selector import (
+    all_reflection_component_selector,
+    round_robin_reflection_component_selector,
+)
 
 
 def optimize(
@@ -166,16 +169,16 @@ def optimize(
     )
 
     if isinstance(module_selector, str):
-        module_selector_cls = {
-            "round_robin": RoundRobinReflectionComponentSelector,
-            "all": AllReflectionComponentSelector,
+        module_selector_fn = {
+            "round_robin": round_robin_reflection_component_selector,
+            "all": all_reflection_component_selector,
         }.get(module_selector)
 
-        assert module_selector_cls is not None, (
+        assert module_selector_fn is not None, (
             f"Unknown module_selector strategy: {module_selector}. Supported strategies: 'round_robin', 'all'"
         )
 
-        module_selector = module_selector_cls()
+        module_selector = module_selector_fn
 
     batch_sampler = EpochShuffledBatchSampler(minibatch_size=reflection_minibatch_size, rng=rng)
 
--- a/src/gepa/api.py
+++ b/src/gepa/api.py
@@ -15,7 +15,7 @@ from gepa.proposer.reflective_mutation.base import LanguageModel, ReflectionComp
 from gepa.proposer.reflective_mutation.reflective_mutation import ReflectiveMutationProposer
 from gepa.strategies.batch_sampler import EpochShuffledBatchSampler
 from gepa.strategies.candidate_selector import CurrentBestCandidateSelector, ParetoCandidateSelector
-from gepa.strategies.component_selector import RoundRobinReflectionComponentSelector
+from gepa.strategies.component_selector import AllReflectionComponentSelector, RoundRobinReflectionComponentSelector
 
 
 def optimize(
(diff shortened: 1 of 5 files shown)


src/gepa/api.py
```diff
@@ -102,7 +102,7 @@ def optimize(
     - perfect_score: The perfect score to achieve.
 
     # Component selection configuration
-    - module_selector: Component selection strategy. Can be a ReflectionComponentSelector instance, a string ('round_robin'), or None. If None, defaults to 'round_robin'. The 'round_robin' strategy cycles through components in order.
+    - module_selector: Component selection strategy. Can be a ReflectionComponentSelector instance, a string ('round_robin', 'all'), or None. If None, defaults to 'round_robin'. The 'round_robin' strategy cycles through components in order. The 'all' strategy selects all components for modification.
```
LakshyAAAgrawal: Let's maybe replace None with 'round_robin'?

LakshyAAAgrawal: "The 'all' strategy selects all components for modification in every GEPA iteration."

src/gepa/strategies/component_selector.py
```diff
@@ -17,8 +17,20 @@ def select_modules(
         candidate: dict[str, str],
     ) -> list[str]:
         pid = state.named_predictor_id_to_update_next_for_program_candidate[candidate_idx]
-        state.named_predictor_id_to_update_next_for_program_candidate[candidate_idx] = (
-            pid + 1
-        ) % len(state.list_of_named_predictors)
+        state.named_predictor_id_to_update_next_for_program_candidate[candidate_idx] = (pid + 1) % len(
```
LakshyAAAgrawal, line 20: This is a weird linting. Can you recheck?

src/gepa/strategies/component_selector.py
```diff
@@ -17,8 +17,20 @@ def select_modules(
...
+class AllReflectionComponentSelector(ReflectionComponentSelector):
+    def select_modules(
+        self,
+        state: GEPAState,
+        trajectories: list[Trajectory],
+        subsample_scores: list[float],
+        candidate_idx: int,
+        candidate: dict[str, str],
+    ) -> list[str]:
+        return list(state.list_of_named_predictors)
```
LakshyAAAgrawal: Perhaps better to return list of candidate keys/

tests/test_module_selector.py
```diff
@@ -73,6 +73,30 @@ def test_module_selector_string_round_robin(mock_proposer, mock_run, common_mock
...
+@patch("gepa.api.GEPAEngine.run")
+@patch("gepa.api.ReflectiveMutationProposer")
+def test_module_selector_string_all(mock_proposer, mock_run, common_mocks):
+    """Test that module_selector='all' works with optimize()."""
+    mock_run_return, mock_adapter = common_mocks
+    mock_run.return_value = mock_run_return
+
+    result = optimize(
+        seed_candidate={"test": "value"},
+        trainset=[],
```
LakshyAAAgrawal: I am surprised this is working with empty trainset!

LakshyAAAgrawal: Can you take a look? Perhaps there is a deeper bug

src/gepa/api.py
andressrg: sure!

src/gepa/strategies/component_selector.py
andressrg: sure

src/gepa/strategies/component_selector.py
andressrg: ah good point

tests/test_module_selector.py
andressrg: Good catch!

It ends up working because the tests are mocking `ReflectiveMutationProposer`, so we are only testing the integration between the compile function and it's arguments.

I agree It's not ideal tho...

src/gepa/strategies/component_selector.py
andressrg: seems to be the pre-commit `ruff-check` 🤔 

I put it back to how it was, ran the pre-commit again and it put it updated it again

src/gepa/strategies/component_selector.py
andressrg: @LakshyAAAgrawal just want to make sure I understood correctly. 

I'm assuming you proposed `candidate.keys()`

It's interesting because even tho they might start the same ( `self.list_of_named_predictors = list(seed_candidate.keys())`). I guess you are referring to the fact that candidate.keys could change! like in this scenario https://github.com/stanfordnlp/dspy/issues/8674

andressrg: I propose we merge this PR, and then we can add a test that verifies what happens if the candidate.keys change

andressrg: Seems to me that test could be easier implemented on the dspy repo

LakshyAAAgrawal:
Did you also have thoughts about #66 ? Maybe we can add that fix here itself. The benefit is that the end user doesn't have to write a class, and a function suffices.

LakshyAAAgrawal:
I was thinking we could make a similar change for the instruction proposer too? Now would be a good time to think about the API changes, before it gets live on a DSPy version.

andressrg:
> Did you also have thoughts about #66 ? Maybe we can add that fix here itself. The benefit is that the end user doesn't have to write a class, and a function suffices.

I like it! Will send the change later today

andressrg:
> Maybe we can add that fix here itself. The benefit is that the end user doesn't have to write a class, and a function suffices.

@LakshyAAAgrawal ready to go 🚀

tests/test_module_selector.py
```diff
@@ -81,14 +114,17 @@ def test_module_selector_custom_instance(mock_proposer, mock_run, common_mocks):
     mock_run.return_value = mock_run_return
 
     class CustomComponentSelector(ReflectionComponentSelector):
-        def select_modules(self, state, trajectories, subsample_scores, candidate_idx, candidate):
+        def __call__(self, state, trajectories, subsample_scores, candidate_idx, candidate):
```
LakshyAAAgrawal: Maybe this can be a function now, not a class?

src/gepa/strategies/component_selector.py
```diff
@@ -17,8 +17,20 @@ def select_modules(
...
-            pid + 1
-        ) % len(state.list_of_named_predictors)
+        state.named_predictor_id_to_update_next_for_program_candidate[candidate_idx] = (pid + 1) % len(
+            state.list_of_named_predictors
+        )
         name = state.list_of_named_predictors[pid]
         return [name]
+
+
+class AllReflectionComponentSelector(ReflectionComponentSelector):
```
LakshyAAAgrawal, line 27: This also can be a function now?

src/gepa/strategies/component_selector.py
```diff
@@ -8,7 +8,7 @@
 
 
 class RoundRobinReflectionComponentSelector(ReflectionComponentSelector):
```
LakshyAAAgrawal, line 10: And this too

LakshyAAAgrawal:
Everything else looks good to me!

andressrg:
Ah good point. Fixing them now

src/gepa/strategies/component_selector.py
```diff
@@ -7,18 +7,33 @@
...
+    subsample_scores: list[float],
+    candidate_idx: int,
+    candidate: dict[str, str],
+) -> list[str]:
+    """Select all components for modification."""
+    return list(candidate.keys())
+
+
+# Explicit declarations that the functions implement the ReflectionComponentSelector protocol
+round_robin_reflection_component_selector: ReflectionComponentSelector = round_robin_reflection_component_selector
```
andressrg: @LakshyAAAgrawal I felt the need to add this to continue declaring explicitly that they follow the ReflectionComponentSelector protocol

but please lmk your thoughts

LakshyAAAgrawal: I am thinking that perhaps we can keep both of these as classes, but with __call__ methods?

LakshyAAAgrawal: The interface allows people to supply methods, but GEPA's internal options for each component is a class.

LakshyAAAgrawal: I think this is the last change. Ready to merge after.

andressrg: sure man

LakshyAAAgrawal:
This is great! Ready to merge?

andressrg:
@LakshyAAAgrawal yes please!

Now AllReflectionComponentSelector and RoundRobinReflectionComponentSelector are classes with __call__ methods. And the ReflectionComponentSelector protocol now makes it easier for users to use plan functions

andressrg:
thanks!

PR #76, opened 2025-09-19
Skip components that are not present in reflective dataset instead of raising KeyError failure

heyalexchoi:
My module architecture has a number of registered submodules that are conditionally routed to.

When ReflectiveMutationProposer tries to retrieve the reflective dataset for one of these submodules, it crashes since there is no trace.

This update handles this case by logging and continuing.

Error details below:
```
2025-09-17 17:33:48,913 - INFO - [MainThread] dspy.teleprompt.gepa.gepa: Iteration 1: Exception during reflection/proposal: 'describe_image_module.predict.predict'
2025-09-17 17:33:48,914 - INFO - [MainThread] dspy.teleprompt.gepa.gepa: Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/gepa/proposer/reflective_mutation/reflective_mutation.py", line 116, in propose
    new_texts = self.propose_new_texts(curr_prog, reflective_dataset, predictor_names_to_update)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/gepa/proposer/reflective_mutation/reflective_mutation.py", line 67, in propose_new_texts
    dataset_with_feedback = reflective_dataset[name]
                            ~~~~~~~~~~~~~~~~~~^^^^^^
KeyError: 'describe_image_module.predict.predict'

2025-09-17 17:33:48,915 - INFO - [MainThread] dspy.teleprompt.gepa.gepa: Iteration 1: Reflective mutation did not propose a new candidate
```

### DIFF
--- a/src/gepa/proposer/reflective_mutation/reflective_mutation.py
+++ b/src/gepa/proposer/reflective_mutation/reflective_mutation.py
@@ -63,6 +63,13 @@ class ReflectiveMutationProposer(ProposeNewCandidate):
 
         new_texts: dict[str, str] = {}
         for name in components_to_update:
+            # Gracefully handle cases where a selected component has no data in reflective_dataset
+            if name not in reflective_dataset or not reflective_dataset.get(name):
+                self.logger.log(
+                    f"Component '{name}' is not in reflective dataset. Skipping."
+                )
+                continue
+
             base_instruction = candidate[name]
             dataset_with_feedback = reflective_dataset[name]
             new_texts[name] = InstructionProposalSignature.run(


LakshyAAAgrawal:
Thanks a lot for the PR @heyalexchoi !

Can you please take a look at https://github.com/gepa-ai/gepa/blob/main/src/gepa/api.py#L37 to see if it handles this scenario already? Basically it allows one to select the components to update given the trajectories, and then the reflection is only called for those traces.

heyalexchoi:
@LakshyAAAgrawal Thanks for GEPA!
If you're referring to `module_selector: "ReflectionComponentSelector | str" = "round_robin",`, it *sort of* does, but:
1. I chose to use the 'all' selector to put off having to implement my own custom selector.
2. The 'all' selector as is, is basically broken if you don't have traces for every named predictor. As soon as it gets to a predictor that doesn't have a trace, it fails for that step. In my case, I came back after a long run to find almost no proposals were made.
3. There's no downside to this guard, and it will help prevent failures in user-developer's custom selectors as well.

If there's anything you think could be improved here, please let me know and I'll update :)

LakshyAAAgrawal:
That makes a lot of sense. Thanks for pointing it out.

In that case, a better approach could be to enable an "all_available" module selector, and add it as an option. The difference will be that, some users may want to treat it as an error if the agent somehow doesn't return traces for all components (who can then continue to use 'all'), and then there are valid usecases like yours that only want to propose new text for the modules which have available traces ('all_available').

PR #81, opened 2025-09-23
Bug Fix: copy base valset eval output as pareto front

aobo-y:
There is a bug that the 1st `val_state` (baseline) of `prog_candidate_val_subscores` in `GEPAState` will be modified once `pareto_front` is updated.

This is coz `base_valset_eval_output[1]` is directly referenced in both `pareto_front_valset` https://github.com/gepa-ai/gepa/blob/main/src/gepa/core/state.py#L45-L52 and `prog_candidate_val_subscores` https://github.com/gepa-ai/gepa/blob/main/src/gepa/core/state.py#L60 

And later `pareto_front_valset`  will be directly modified with new scores in place https://github.com/gepa-ai/gepa/blob/main/src/gepa/core/state.py#L131

so just make a copy

### DIFF
--- a/src/gepa/core/state.py
+++ b/src/gepa/core/state.py
@@ -42,7 +42,7 @@ class GEPAState(Generic[RolloutOutput]):
         track_best_outputs: bool = False,
     ):
         valset_base_score = sum(base_valset_eval_output[1]) / len(base_valset_eval_output[1])
-        base_valset_pareto_front = base_valset_eval_output[1]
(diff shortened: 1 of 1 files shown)


LakshyAAAgrawal:
Thanks a lot for pointing this out and providing a fix @aobo-y!

LakshyAAAgrawal:
I am wondering if there's some neat way to capture this bug as a unit test, that would fail before the fix through this PR, but pass now, that we can add to GEPA unit tests. Would ensure that we don't reintroduce such bugs!

PR #85, opened 2025-09-24
feature: add testset to GEPA engine

Shangyint:
TODO:
make testset optional (and should be none by default)


LakshyAAAgrawal:
I think a more general formulation would allow adding multiple datasets, which are just tracking_datasets. So instead of calling it testset, we can call it "tracking_dataset"

Shangyint:
closing thi

Issue #56, opened 2025-09-04
Restart from log_dir not working


LakshyAAAgrawal:

enitrat:
Duplicate from https://github.com/gepa-ai/gepa/issues/42 ?

Issue #42, opened 2025-08-29
Allow GEPA to continue running from a partially completed run


LakshyAAAgrawal:

enitrat:
working on this

PR #86, opened 2025-09-24
fix: gepa state initialization from run_dir

enitrat:
- fixes state initialization from `run_dir` by removing condition on `prog_candidates` path existence
- added associated tests ensuring that GEPA.save --> initialize works as expected

Closes https://github.com/gepa-ai/gepa/issues/56
Closes https://github.com/gepa-ai/gepa/issues/42


enitrat:
hey @LakshyAAAgrawal anything that needs to be done regarding the CI failure? Seems unrelated

LakshyAAAgrawal:
Dear @TomeHirata, could you please have a look at (https://github.com/gepa-ai/gepa/actions/runs/17974503472/job/51158303766?pr=86#step:8:218). I think the experiment tracker tests are flaky, as I have seen this fail in some other CI runs too, but typically goes away by rerunning the jobs.

LakshyAAAgrawal:
@enitrat , this is not a necessary requirement to merge this PR, but do you think you would be able to add an end to end gepa workflow testing the save state feature? Something similar to (https://github.com/gepa-ai/gepa/blob/main/tests/test_aime_prompt_optimization/test_aime_prompt_optimize.py), but it runs GEPA once for N steps with run_dir, and after it returns, calls gepa again for N steps, and the second time it returns, it should return results equivalent to 2*N steps. It would be ideal if we can capture this in the test, but no worries if you can't do it in this PR.

enitrat:
im not sure how to properly harness an e2e test by i'll look into it. in the meanwhile the targeted unit tests should have basic coverage

LakshyAAAgrawal:
The file I linked provides a fully deterministic harness (just uses a pre-recorded dictionary). Hopefully that is helpful!

enitrat:
@LakshyAAAgrawal here's my proposal:

1. moved some fixtures to a root conftest for shared usage
2. created a specific tests/cache directory related to shared fixtures
3. for the e2e test, first run with max_metric_calls=N, then verify when called with max_metric_calls=0 on second run, we have the same amount of metric calls as first run (meaning the state has been properly loaded.)

tests/test_state.py
```diff
@@ -0,0 +1,132 @@
...
+    assert result.num_full_ds_evals == 1
+    assert result.total_num_evals == len(valset_out[1])
+    fake_logger.log.assert_not_called()
+    valset_evaluator.assert_called_once_with(seed)
+
+    # Files written for each task with outputs (not scores)
+    base = run_dir / "generated_best_outputs_valset"
+    p0 = base / "task_0" / "iter_0_prog_0.json"
+    p1 = base / "task_1" / "iter_0_prog_0.json"
+    assert p0.exists() and p1.exists()
```
LakshyAAAgrawal, line 44: The test should clean up the run_dir. Is that happening somewhere?

LakshyAAAgrawal:
Dear @enitrat, this is a great refactor of the tests. Thank you very much!

Can you just look into the cleanup of the run_dir arguments?

tests/test_state.py
enitrat: no, because `run_dir` is created using the [`tmp_path` fixture](https://docs.pytest.org/en/stable/how-to/tmp_path.html#the-tmp-path-fixture) which is temporary to each test function and not retained

tests/cache/llm_cache.json
LakshyAAAgrawal, line 1: Can you move this back to where it was? Right now the distance between llm_cache and the place it is used has increased.

LakshyAAAgrawal: Question: How can the updated fixture support different record dirs?

enitrat: > Can you move this back to where it was? Right now the distance between llm_cache and the place it is used has increased.

I can do that but considering that it's no longer only used for aime_prompt_optimmization but all kinds of e2e tests, it felt better in a shared, common directory

> Question: How can the updated fixture support different record dirs?

the `mocked_lms` takes as argument a `recorder_dir` which dictates where to save the cache. 

There is a "default, root" `recorder_dir()` fixture that exists in `conftests.py`, but you can define local ones closer to the test modules that will take over priority over the conftest one.

LakshyAAAgrawal: I think it would still be better to keep it at aime_prompt_op. The reason is that it is the place where the recording was performed. This way, a new reader can easily figure out where a specific recorded cache comes from.

enitrat: done 👍

LakshyAAAgrawal:
This is a great PR. Thanks a lot @enitrat!

TomeHirata:
@LakshyAAAgrawal Yep, filed https://github.com/gepa-ai/gepa/pull/89

PR #94, opened 2025-10-03
Use cloudpickle if available.

EthanAtLL:
I'd like to use `dspy.GEPA` with the `log_dir` parameter for checkpointing, but [DSPy Issue 1221](https://github.com/stanfordnlp/dspy/issues/1221) stops me. As I understand it the pickle is unable to serialize the [dynamically created class created by SignatureMeta](https://github.com/stanfordnlp/dspy/blob/965da1db9cec611710534d42b9049fdbc13d88d7/dspy/signatures/signature.py#L49).   To get around issue, this MR replaces pickle with `cloudpickle` when the it can be imported and otherwise falls back to `pickle` in `src/gepa/core/state.py` so the dynamic class can be serialized. No dependency was added.

### DIFF
--- a/src/gepa/core/state.py
+++ b/src/gepa/core/state.py
@@ -86,14 +86,20 @@ class GEPAState(Generic[RolloutOutput]):
         if run_dir is None:
             return
         with open(os.path.join(run_dir, "gepa_state.bin"), "wb") as f:
-            import pickle
+            try:
+                import cloudpickle as pickle
+            except ImportError:
+                import pickle
             d = dict(self.__dict__.items())
             pickle.dump(d, f)
 
     @staticmethod
     def load(run_dir: str) -> "GEPAState":
         with open(os.path.join(run_dir, "gepa_state.bin"), "rb") as f:
-            import pickle
+            try:
+                import cloudpickle as pickle
+            except ImportError:
+                import pickle
             d = pickle.load(f)
         state = GEPAState.__new__(GEPAState)
         state.__dict__.update(d)


LakshyAAAgrawal:
Dear @EthanAtLL,

Thank you very much for this PR.

Question: What happens when cloudpickle is available, but I am trying to deserialize a file created with pickle? Can we have an in-built mechanism that detects which serialization was used?

EthanAtLL:
I believe that's ok as cloudpickle uses the standard [Python unpickler](https://github.com/cloudpipe/cloudpickle/blob/master/cloudpickle/cloudpickle.py#L17),  cloudpickle.load() defers to [pickle.load()](https://github.com/cloudpipe/cloudpickle/blob/master/cloudpickle/cloudpickle.py#L1539).

LakshyAAAgrawal:
Hi @EthanAtLL, I think we should have a flag to determine if cloudpickle is used or not. From my reading about cloudpickle, it can lead to performance degradation for certain object classes, and may have more compatibility and versioning issues (across python versions and environments).

Having said that, it is clearly very useful for situations like the above. Hence, this should be a configurable flag, that one can pass to gepa.optimize (`use_cloudpickle: bool = False`).

pyproject.toml
```diff
@@ -25,7 +25,8 @@ full = [
     "datasets>=2.14.6",
     "mlflow>=3.0.0",
     "wandb",
-    "tqdm>=4.66.1"
+    "tqdm>=4.66.1",
+    "cloudpickle>=3.1.1",
```
LakshyAAAgrawal: I would recommend avoiding a new dependency for now.

LakshyAAAgrawal (dismissed):
Everything looks good, except for adding new dependency.

pyproject.toml
EthanAtLL: Sure, rebased dropping the pyproject.toml and uv.lock updates.

LakshyAAAgrawal:
Thank you very much for the PR, @EthanAtLL!

PR #98, opened 2025-10-07
Fix reflective proposer stalling by consuming metric budget on trace failures

### LINKED ISSUE #46: Non-progress state in GEPA’s optimization loop?
The GEPA loop sometimes appears to “stall” (iterations continue (log files explode in size!) while the active termination budget doesn’t advance.

### What I think might be happening

* **Single active budget**: Only one of `num_iters`, `max_evals_per_trainval_instance`, or `max_metric_calls` is set; the others are `None`, so the `while` guard stays lenient.
* **Unusable feedback**: `capture_module_trace_with_feedback(...)` returns `(None, None, None)` (e.g., parse failures when `add_format_failure_as_feedback=False`, or all examples perfect with `skip_perfect_score=True`).
* **Counters increment only on success**: On unusable feedback, the loop `continue`s before any budget counter is incremented.
* **Repeated failures**: If unusable feedback persists, `gepa_state.i` increases, but the active budget doesn’t, so the `while` condition never trips.

### Code path I’m reading

```python
while (
    (self.num_iters is None or gepa_state.num_full_ds_evals < self.num_iters)
    and ...
    and ...
):
    dataset_with_feedback, subsample_score, subsample_scores = (
        capture_module_trace_with_feedback(...)
    )

    if dataset_with_feedback is None or subsample_score is None:
        self.logger.log("No feedback samples, skipping")
        continue  # possible skip of any budget increment
    # ...
    gepa_state.total_num_evals += len(subsample_ids)  # runs only on success
```

fgn:
Increment `state.total_num_evals` immediately after the minibatch evaluation, even when no trajectories are captured. Without this, the proposer returned early before updating the counter, so MaxMetricCallsStopper (which checks
  `total_num_evals >= max_metric_calls`) never triggered and the optimization loop could run indefinitely. 

This change restores the budget accounting and prevents the infinite loop.

Fixes #46

### DIFF
--- a/src/gepa/proposer/reflective_mutation/reflective_mutation.py
+++ b/src/gepa/proposer/reflective_mutation/reflective_mutation.py
@@ -92,12 +92,12 @@ class ReflectiveMutationProposer(ProposeNewCandidate):
 
         # 1) Evaluate current program with traces
         eval_curr = self.adapter.evaluate(minibatch, curr_prog, capture_traces=True)
+        state.total_num_evals += len(subsample_ids)
+        state.full_program_trace[-1]["subsample_scores"] = eval_curr.scores
+
         if not eval_curr.trajectories or len(eval_curr.trajectories) == 0:
             self.logger.log(f"Iteration {i}: No trajectories captured. Skipping.")
             return None
-
-        state.total_num_evals += len(subsample_ids)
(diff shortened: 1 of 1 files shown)


LakshyAAAgrawal (approved):

LakshyAAAgrawal:
Thank you so much for the PR!

PR #99, opened 2025-10-07
Fix parsing for ``` blocks with language in InstructionProposalSignature

CDBiddulph:
This allows us to properly parse code blocks written by the reflection LM when they include a language specifier.

For example, this "plaintext" block:
```plaintext
These are the instructions.
```
would previously parse to "plaintext\nThese are the instructions." After this change, it will parse to "These are the instructions."

### DIFF
--- a/src/gepa/strategies/instruction_proposal.py
+++ b/src/gepa/strategies/instruction_proposal.py
@@ -1,6 +1,8 @@
 # Copyright (c) 2025 Lakshya A Agrawal and the GEPA contributors
 # https://github.com/gepa-ai/gepa
 
+import re
+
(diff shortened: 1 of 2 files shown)


LakshyAAAgrawal:
Hi @CDBiddulph, I wanted to ensure that we capture the largest/last block returned by the LLM. Can you confirm this ensures that?

CDBiddulph:
Ah, I see, the original code would find the "outermost" block. I fixed it to work this way. You can see the test cases to check that this looks correct.

LakshyAAAgrawal:
This is great! Thank you very much

Issue #96, opened 2025-10-06
Can we add new evals?


LakshyAAAgrawal:
Currently, adding new evals to the trainset is supported for cumulative runs via log_dir, which reuses optimization history. However, adding to the valset is not yet supported. This would be a trivial change and align with plans for a more flexible system where not all prompt candidates need evaluation on all valset instances.

Also support not having to run full valset evaluation for every new program discovered.

aria42:
Hey Lakshy, I'm happy to take a stab writing a PR for this! This is a really great project and I think I've got enough handle on the internals to take a stab.

vincentkoc:
Can this be closed? assumed valset is fully supported?

PR #100, opened 2025-10-08
Dynamic Validation in GEPAState

aria42:
Addresses #96 and enables add new validation evals. Also removes assumption that every program has been evaluated on all validation examples. Everything is still well-defined with this change with the exception that merge proposals must be careful to compare only on validation examples both candidates have been evaluated against.

## Changes
* Changes to `GEPAState`: The state information about validation (scores, outputs, etc.) are stored in terms of opaque identifier (`ValId`). Making the state agnostic about the set of validation examples enables you to extend the validation examples from existing states (and run directories). It also removes the assumption that we have "dense" evaluation where each program is evaluated on the full validation set. 

* Changes to `GEPAEngine`: The engine takes optional `val_evaluation_policy` to select the subset of validation examples to evaluate for a program. If none is provided the full valset is evaluated. The policy enables you to have flexible ways of choosing examples (e.g., picking validation examples not evaluated by other programs or curriculum-style staging of more harder examples over time). Note that `GEPAEngine` assumes the valset ids are indices of a validation set (which can be updated), but this assumption doesn't leak to `GEPAState`. 

* Changes to `api.py`:  Add a parameter for `merge_val_overlap_floor` which is threaded through to `MergeProposer`. This parameter rejects merge proposal if the merge candidates don't have this many shared validation examples they are both evaluated against.

## Tests
Existing tests pass and added the following:
* Added test around extending validation and having custom validation selection policy 
* Added test around loading a legacy `gepa.bin` format and validating it can be read and upgraded to "sparse" format

## Open Questions
* In `GEPAState` there was `per_program_tracked_scores` and `program_full_scores_val_set` member variables which were meant to capture the train validation and valset scores for candidates. But prior to this change these two variables were identical (latter was set to the former see existing code [here](https://github.com/gepa-ai/gepa/blob/ba06c3bf898d66eeb7d724b0aabaf19272956a4a/src/gepa/core/state.py#L127-L128) and [here](https://github.com/gepa-ai/gepa/blob/ba06c3bf898d66eeb7d724b0aabaf19272956a4a/src/gepa/core/state.py#L148)). I've left derived properties for these for backward compatibility but could remove them since I think the distinction is no longer needed.


src/gepa/proposer/merge.py
```diff
@@ -260,6 +290,11 @@ def propose(self, state: GEPAState) -> CandidateProposal | None:
             state.prog_candidate_val_subscores[id1],
             state.prog_candidate_val_subscores[id2],
         )
+        if not subsample_ids:
```
LakshyAAAgrawal: Question: I do several random selections to select id1 and id2 to see if we can find candidates which we are able to merge that also satisfy all the requirements. Do you think we can include this as part of those checks?

src/gepa/logging/utils.py
```diff
@@ -14,19 +14,26 @@ def log_detailed_metrics_after_discovering_new_program(
...
-    logger.log(f"Iteration {gepa_state.i + 1}: Individual valset scores for new program: {valset_subscores}")
-    logger.log(f"Iteration {gepa_state.i + 1}: New valset pareto front scores: {gepa_state.pareto_front_valset}")
     logger.log(
-        f"Iteration {gepa_state.i + 1}: Full valset pareto front score: {sum(gepa_state.pareto_front_valset) / len(gepa_state.pareto_front_valset)}"
+        f"Iteration {gepa_state.i + 1}: Train/val aggregate for new program: {gepa_state.per_program_tracked_scores[new_program_idx]}"
     )
+    logger.log(f"Iteration {gepa_state.i + 1}: Individual valset scores for new program: {valset_subscores}")
+    logger.log(f"Iteration {gepa_state.i + 1}: New valset pareto front scores: {gepa_state.pareto_front_valset}")
+
+    pareto_scores = [score for score in gepa_state.pareto_front_valset.values() if score != float("-inf")]
```
LakshyAAAgrawal: I am wondering if we should enforce that at least the first program, or some program is evaluated on the full valset, just so we have a coverage of all the problems?

src/gepa/gepa_utils.py
```diff
@@ -54,36 +54,43 @@ def remove_dominated_programs(program_at_pareto_front_valset, scores=None):
...
         assert any(p in front for p in dominators)
 
-    new_program_at_pareto_front_valset = [{prog_idx for prog_idx in front if prog_idx in dominators} for front in program_at_pareto_front_valset]
-    assert len(new_program_at_pareto_front_valset) == len(program_at_pareto_front_valset)
-    for front_old, front_new in zip(program_at_pareto_front_valset, new_program_at_pareto_front_valset, strict=False):
-        assert front_new.issubset(front_old)
+    new_program_at_pareto_front_valset = {
+        val_id: {prog_idx for prog_idx in front if prog_idx in dominators}
+        for val_id, front in program_at_pareto_front_valset.items()
+    }
```
LakshyAAAgrawal, line 71: The current candidate selection logic assumes that all valset scores are present. I expect the current code will break if this assumption is not satisfied.

It could perhaps be good to encode this as an assert statement somewhere in https://github.com/gepa-ai/gepa/blob/main/src/gepa/strategies/candidate_selector.py#L11-L32.

src/gepa/gepa_utils.py
```diff
@@ -54,36 +54,43 @@ def remove_dominated_programs(program_at_pareto_front_valset, scores=None):
...
+    for testcase_pareto_front in new_program_at_pareto_front_valset.values():
         for prog_idx in testcase_pareto_front:
             if prog_idx not in program_frequency_in_validation_pareto_front:
                 program_frequency_in_validation_pareto_front[prog_idx] = 0
             program_frequency_in_validation_pareto_front[prog_idx] += 1
 
     sampling_list = [prog_idx for prog_idx, freq in program_frequency_in_validation_pareto_front.items() for _ in range(freq)]
-    assert len(sampling_list) > 0
+    if not sampling_list:
+        return idxmax(train_val_weighted_agg_scores_for_all_programs)
```
LakshyAAAgrawal: Will idxmax work on "train_val_weighted_agg_scores_for_all_programs" directly?

src/gepa/api.py
```diff
@@ -57,6 +59,7 @@ def optimize(
     # Reproducibility
     seed: int = 0,
     raise_on_exception: bool = True,
+    val_evaluation_policy: Callable[[GEPAState], list[int]] | None = None,
```
LakshyAAAgrawal: Tying back to the point about some candidate selectors not working with partial validation scores, maybe in the gepa.optimize call (entrypoint), we can add the assert here. That if val_evaluation_policy is something that doesn't work with the candidate selector chosen, then throw an error.

LakshyAAAgrawal:
Could we also address #79, which would be very aligned with this PR? It should also be a straightforward change. Just need to implement a DataLoader protocol, and load next training minibatch using it.

src/gepa/core/engine.py
```diff
@@ -44,6 +44,7 @@ def __init__(
         raise_on_exception: bool = True,
         # Budget and Stop Condition
         stop_callback: Callable[[Any], bool] | None = None,
+        val_evaluation_policy: Callable[[GEPAState], list[int]] | None = None,
```
LakshyAAAgrawal: We can create a python Protocol for val_evaluation_policy.

src/gepa/core/engine.py
```diff
@@ -88,22 +101,24 @@ def _run_full_eval_and_add(
     ) -> tuple[int, int]:
         num_metric_calls_by_discovery = state.total_num_evals
 
-        valset_outputs, valset_subscores = self._val_evaluator()(new_program)
-        valset_score = sum(valset_subscores) / len(valset_subscores)
+        valset_outputs, valset_subscores = self._evaluate_on_valset(new_program, state)
+        valset_score = (
+            sum(valset_subscores.values()) / len(valset_subscores) if len(valset_subscores) > 0 else float("-inf")
```
LakshyAAAgrawal: nit: Any thoughts about float(-inf) vs NaN?

LakshyAAAgrawal:
Any thoughts about the scenario where we have an updating/expanding valset? I think the current valset still assumes in some places that the size of the valset is fixed. Since we are generalizing, we could perhaps try to aim for the most general case. 

I like the concept of ValId introduced as a typevar. I think this enables us to expand validation sets?

src/gepa/core/state.py
```diff
@@ -3,21 +3,30 @@
...
+ValId = TypeVar("ValId", bound=Hashable)
+"""Opaque identifier for valset examples"""
+ValScores: TypeAlias = dict[ValId, float]
+ValOutputs: TypeAlias = dict[ValId, RolloutOutput]
 
-class GEPAState(Generic[RolloutOutput]):
-    program_candidates: list[dict[str, str]]
-    parent_program_for_candidate: list[list[int | None]]
 
-    program_full_scores_val_set: list[float]
```
LakshyAAAgrawal, line 16: I think we should still keep the aggregate scores for all programs? I am trying to think of arguments for both ways. What are your thoughts?

src/gepa/core/state.py
```diff
@@ -3,21 +3,30 @@
...
+class GEPAState(Generic[RolloutOutput, ValId]):
+    _VALIDATION_SCHEMA_VERSION: ClassVar[int] = 2
```
LakshyAAAgrawal, line 21: Should we introduce something like this for GEPAResult class as well?

src/gepa/core/state.py
```diff
@@ -30,55 +39,50 @@ class GEPAState(Generic[RolloutOutput]):
...
+        self.program_at_pareto_front_valset = {val_id: {0} for val_id in base_scores.keys()}
 
         self.list_of_named_predictors = list(seed_candidate.keys())
         self.named_predictor_id_to_update_next_for_program_candidate = [0]
         self.i = -1
 
-        self.prog_candidate_val_subscores = [base_valset_eval_output[1]]
         self.num_metric_calls_by_discovery = [0]
 
-        if track_best_outputs:
```
LakshyAAAgrawal, line 63: Why delete this? The main reason this exists is that for some tasks, the output can be very large, and consume a lot of ram. So unless the user really cares about it, we offer the flexibility to garbage collect this by not storing.

src/gepa/core/state.py
```diff
@@ -87,84 +91,164 @@ def save(self, run_dir: str | None):
...
+        assert set(state.pareto_front_valset.keys()) == set(state.program_at_pareto_front_valset.keys())
+        assert len(state.program_candidates) == len(state.prog_candidate_val_subscores)
+        assert len(state.program_candidates) == len(state.num_metric_calls_by_discovery)
         assert len(state.program_candidates) == len(state.parent_program_for_candidate)
         assert len(state.program_candidates) == len(state.named_predictor_id_to_update_next_for_program_candidate)
-
         return state
 
+    @staticmethod
+    def _migrate_legacy_state_dict(d: dict[str, Any]) -> None:
```
LakshyAAAgrawal: Should we introduce a notion of schema versions here?

if loaded version == NaN OR loaded version == 1:
    migrate_1()
elif loaded version == 2:
    migrate_2()

src/gepa/core/state.py
```diff
@@ -87,84 +91,164 @@ def save(self, run_dir: str | None):
...
+        if isinstance(program_at_front, list):
+            d["program_at_pareto_front_valset"] = {idx: set(front) for idx, front in enumerate(program_at_front)}
+
+        best_outputs = d.get("best_outputs_valset")
+        if isinstance(best_outputs, list):
+            d["best_outputs_valset"] = {idx: list(outputs) for idx, outputs in enumerate(best_outputs)}
+
+        d["validation_schema_version"] = GEPAState._VALIDATION_SCHEMA_VERSION
+
+    def get_program_average(self, program_idx: int) -> tuple[float, int]:
```
LakshyAAAgrawal: We should call this get_program_average_val_subset() or something like that, since it doesn't represent full valset score.

src/gepa/core/state.py
```diff
@@ -87,84 +91,164 @@ def save(self, run_dir: str | None):
...
+            for val_id in val_scores.keys():
+                result[val_id].append(program_idx)
+        return result
+
+    @property
+    def program_full_scores_val_set(self) -> list[float]:
+        return [self.get_program_average(program_idx)[0] for program_idx in range(len(self.prog_candidate_val_subscores))]
+
+    @property
+    def per_program_tracked_scores(self) -> list[float]:
```
LakshyAAAgrawal: We can delete this now. It is an artifact of an internal version of the code I had months before release.

src/gepa/core/state.py
```diff
@@ -87,84 +91,164 @@ def save(self, run_dir: str | None):
...
+            output = outputs.get(val_id) if outputs is not None else None
+            if self.best_outputs_valset is not None and output is not None:
+                self.best_outputs_valset[val_id] = [(program_idx, output)]
+                if run_dir is not None:
+                    task_dir = os.path.join(run_dir, "generated_best_outputs_valset", f"task_{val_id}")
+                    os.makedirs(task_dir, exist_ok=True)
+                    with open(os.path.join(task_dir, f"iter_{iteration}_prog_{program_idx}.json"), "w") as fout:
+                        json.dump(output, fout, indent=4, default=json_default)
+        elif score == prev_score:
+            pareto_front = self.program_at_pareto_front_valset.setdefault(val_id, set())
```
LakshyAAAgrawal: There can actually be an assert here, since the score == prev_score, there must be a non-empty set there already.

src/gepa/core/state.py
```diff
@@ -87,84 +91,164 @@ def save(self, run_dir: str | None):
...
-                    self.best_outputs_valset[task_idx].append((new_program_idx, valset_outputs[task_idx]))
+        for val_id, score in valset_subscores.items():
+            self._update_pareto_front_for_val_id(val_id, score, new_program_idx, valset_outputs, run_dir, self.i + 1)
 
-        assert len(valset_subscores) == len(self.program_at_pareto_front_valset)
+        linear_pareto_front_program_idx = self._best_program_idx()
+        return new_program_idx, linear_pareto_front_program_idx
 
-        self.per_program_tracked_scores = self.program_full_scores_val_set
+    def _best_program_idx(self) -> ProgramIdx:
```
LakshyAAAgrawal: Perhaps this needs to come from the val_evaluation_policy itself? That protocol should provide both a way to select validation instances, and also to select the best program given that instance too.

src/gepa/core/state.py
```diff
@@ -87,84 +91,164 @@ def save(self, run_dir: str | None):
...
 
-        linear_pareto_front_program_idx = idxmax(self.per_program_tracked_scores)
 
-        return new_program_idx, linear_pareto_front_program_idx
+def write_eval_scores_to_directory(scores: ValScores, output_dir: str):
+    for val_id, score in scores.items():
+        task_dir = os.path.join(output_dir, f"task_{val_id}")
+        os.makedirs(task_dir, exist_ok=True)
+        with open(os.path.join(task_dir, f"iter_{0}_prog_0.json"), "w") as f:
+            json.dump(score, f, indent=4, default=json_default)
```
LakshyAAAgrawal, line 241: This line was earlier writing both the score and output I think.

tests/test_module_selector.py
```diff
@@ -16,8 +16,8 @@ def common_mocks():
         program_candidates=[{"test": "value"}],
         parent_program_for_candidate=[None],
         program_full_scores_val_set=[0.5],
-        prog_candidate_val_subscores=[[]],
-        program_at_pareto_front_valset=[set()],
+        prog_candidate_val_subscores=[{}],
+        program_at_pareto_front_valset={},
```
LakshyAAAgrawal: Shouldn't the strict conversion be `{0: set()}`?

LakshyAAAgrawal:
Dear @aria42,

Thank you so much for this wonderful PR. This is exactly the direction I wanted to take GEPA towards, and I am hoping that merging this will enable deep research into ideal, data and rollout efficient validation as well as candidate selection strategies. Thank you once again for the PR.

SInce the PR changes so many parts of the codebase, I have left a lot of comments. Please feel free to ignore some of them if you disagree, and I would love to hear your thoughts about it.

aria42:
> SInce the PR changes so many parts of the codebase, I have left a lot of comments. Please feel free to ignore some of them if you disagree, and I would love to hear your thoughts about it.

Thanks @LakshyAAAgrawal! Comments are great will address in the morning.

src/gepa/core/state.py
aria42: It's converted into a property which is derived from the raw data on the fly below. You have another comment to rename this which I'll do.  I think keeping as a derived property is fine as long as the data sizes here are pretty small and we're not worried about the cost of on-the-fly aggregation right now.

src/gepa/core/state.py
aria42: This function is basically `migrate_1` but I'll make it clearer. The callsite for this is where I meant there to be dispatching upcasting migration depending on the version.

src/gepa/core/state.py
aria42: ack

README.md
```diff
@@ -39,9 +39,12 @@ pip install git+https://github.com/gepa-ai/gepa.git
 GEPA can be run in just a few lines of code. In this example, we'll use GEPA to optimize a system prompt for math problems from the AIME benchmark ([full tutorial](https://dspy.ai/tutorials/gepa_aime/)). Run the following in an environment with `OPENAI_API_KEY`:
 ```python
 import gepa
+from gepa.core.data_loader import ListDataLoader
```
LakshyAAAgrawal: Can we ensure that the end user's experience is not changed. For example, in this scenario, gepa.optimize(trainset: List | DataLoader), and that way, the old interface still works, and inside gepa.optimize, we can see if the user just directly provides lists, then it uses ListDataLoader.

LakshyAAAgrawal: Also, how would ListDataLoader tie with https://github.com/gepa-ai/gepa/blob/main/src/gepa/strategies/batch_sampler.py#L10?

BatchSampler basically is the data selection strategy for trainset.

src/gepa/api.py
```diff
@@ -25,8 +27,8 @@
 
 def optimize(
     seed_candidate: dict[str, str],
-    trainset: list[DataInst],
-    valset: list[DataInst] | None = None,
+    trainset: DataLoader[DataId, DataInst],
```
LakshyAAAgrawal: This should be `List | DataLoader[DataId, DataInst]`

src/gepa/core/adapter.py
```diff
@@ -4,10 +4,12 @@
 from dataclasses import dataclass
 from typing import Any, Generic, Protocol, TypeVar
 
+from .data_loader import DataInst
+
 # Generic type aliases matching your original
 RolloutOutput = TypeVar("RolloutOutput")
 Trajectory = TypeVar("Trajectory")
-DataInst = TypeVar("DataInst")
```
LakshyAAAgrawal, line 10: Since DataInst touches the adapter, which is the core user-facing abstraction from GEPA, it should be close to adapter's definition imo.

src/gepa/examples/anymaths-bench/train_anymaths.py
```diff
@@ -159,8 +160,8 @@ def reflection_lm(prompt: str):
 
     optimized_results = optimize(
         seed_candidate={"instruction_prompt": seed_instruction},
-        trainset=trainset,
-        valset=valset,
+        trainset=ListDataLoader(trainset),
```
LakshyAAAgrawal: Once the gepa.optimize can take both a list and this, then this can be undo.

src/gepa/examples/rag_adapter/rag_optimization.py
```diff
@@ -752,8 +753,8 @@ def main():
 
             result = gepa.optimize(
                 seed_candidate=initial_prompts,
-                trainset=train_data,
-                valset=val_data,
+                trainset=ListDataLoader(train_data),
```
LakshyAAAgrawal: Same

src/gepa/examples/terminal-bench/train_terminus.py
```diff
@@ -132,8 +133,8 @@ def perform_task(
 
     optimized_results = optimize(
         seed_candidate={"instruction_prompt": initial_prompt_from_terminus},
-        trainset=trainset,
-        valset=valset,
+        trainset=ListDataLoader(trainset),
```
LakshyAAAgrawal: Same

src/gepa/core/engine.py
aria42: Slight preference  for -inf so comparisons with real numbers are still coherent, but a weak preference and happy to go with NaN if you prefer

src/gepa/core/state.py
aria42: Added to `EvaluationPolicy` protocol. I have some open question about what should be in this protocol in the comment below, but this specific method makes sense.

src/gepa/core/state.py
aria42: I don't think it was. I thought so as well, but [the unit tests](https://github.com/gepa-ai/gepa/blob/main/tests/test_state.py#L41-L46) for this read these json files and check for just a score. I'm fine to have this dump both (which makes sense!) but wasn't sure what the intended behavior was @LakshyAAAgrawal

src/gepa/gepa_utils.py
aria42: The only current [call site](https://github.com/gepa-ai/gepa/blob/main/src/gepa/strategies/candidate_selector.py#L20-L23) for this passes in `state.program_full_scores_val_set` is a dense array of average validation scores for each program. So `idxmax` should work, but of course this PR changes the semantics so you're comparing averages that <I>could</I> be from different validation examples.

src/gepa/logging/utils.py
aria42: So this is actually the case right now since we call `initialize_gepa_state` when we start the engine and there we run the seed candidate against the full validation set, so the seed candidate should provide some score for all examples. I've removed the -inf checks here since they aren't needed and converted to asserts

README.md
aria42: Addressed this in 0aefafa9

aria42:
Hey @LakshyAAAgrawal,

I've got a few items open (the unresolved comments), but wanted to update on some stuff to get your opinion. I can also squash my full diff into a single commit since merging the main commits has made this PR hard to read now 😄 

- Check out `eval_policy.py` this is the "validation eval policy"  protocol that selects the validation examples to use for evaluation. With a tweak to `batch_sampler.py` it can re-use the same protocol used for training batch selection. It extends this with a method to get `get_best_program` on validation (to handle how to compare programs with different eval support coverage). I thought about also having `get_best_program_for_instance` to have a per-instance level example, but this gets a little messy in terms of interaction with the existing Pareto front logic.

- I believe the interface for `api.py` and `engine.py` is now fully backward compatible since we accept lists or data loader for train/validation. 

I still need to address open comments but big things left to tackle in morning
[x] Validating the candidate selector is compatible with the eval policy 
[x] Pushing the check for sufficient overlap for merge proposals deeper into logic checks for selecting candidates
[x] Add more unit test coverage for changes

src/gepa/proposer/merge.py
aria42: I pushed this deeper into the merge code while you're scanning for valid pairs by injecting a predicate. I also added some tests on the merge functions since I wasn't 100% sure on the types everywhere in 889e1a9f

src/gepa/gepa_utils.py
aria42: I'm not sure this is the best solution, but basically the `CandidateSelector` is responsible for saying if it's compatible with an `EvaluationPolicy` and the Pareto selector throws if the eval policy is "sparse".  See 0d651a

src/gepa/api.py
aria42: Handled in 0d651a and added a test case that this fails

aria42:
Hey @LakshyAAAgrawal,

I think I've covered the comments (but easily could've missed one). I'm not thrilled about the `EvaluationPolicy` interface; I think making it a `BatchSampler` (which is similar) makes the names a bit confusing. I think it all works but might be better to separate. 

Let me know, totally happy to iterate more. And sorry the PR got so big, this just ended up touching lots of things.

src/gepa/core/state.py
aria42: Addressed in 6712d4

src/gepa/core/state.py
LakshyAAAgrawal: Let's make a flag, that determines if the output will be written or not. One of the usecases of GEPA is as an inference-time scaling technique, where the ability to dump outputs will be very valuable.

src/gepa/api.py
```diff
@@ -58,6 +62,7 @@ def optimize(
     # Reproducibility
     seed: int = 0,
     raise_on_exception: bool = True,
+    val_evaluation_policy: EvaluationPolicy[DataId, DataInst] | None = None,
```
LakshyAAAgrawal: This should be `EvaluationPolicy[DataId, DataInst] | str | None` following other arguments, where the user can provide names like "full_val_eval" or something like that

src/gepa/api.py
```diff
@@ -114,6 +119,7 @@ def optimize(
     # Merge-based configuration
     - use_merge: Whether to use the merge strategy.
     - max_merge_invocations: The maximum number of merge invocations to perform.
+    - merge_val_overlap_floor: Minimum number of shared validation ids required between parents before attempting a merge subsample.
```
LakshyAAAgrawal: Add: "Only relevant when using an eval policy different from 'full'"

src/gepa/gepa_utils.py
LakshyAAAgrawal: Can you add a comment above this line making note of this.

src/gepa/core/engine.py
```diff
@@ -74,10 +79,21 @@ def __init__(
...
+        self.val_evaluation_policy = val_evaluation_policy
 
-    def _val_evaluator(self) -> Callable[[dict[str, str]], tuple[list[RolloutOutput], list[float]]]:
+    def _evaluate_on_valset(
+        self, program: dict[str, str], state: GEPAState
+    ) -> tuple[dict[int, RolloutOutput], dict[int, float]]:
         assert self.valset is not None
-        return lambda prog: self.evaluator(self.valset, prog)
+
+        val_ids = self.val_evaluation_policy.next_minibatch_ids(self.valset, state)
```
LakshyAAAgrawal: Can we call this something other than minibatch? minibatch is already being used as a term to refer to the training/reflection minibatch size so could be confusing.

LakshyAAAgrawal:
@aria42 I think since the goal of the PR is to introduce dynamic valset, it would be useful to have at least one example run that actually uses a dynamic valset. That example could be included as an e2e test, or as a runnable notebook example. Wdyt? Implementing an evaluation_policy that is different from the "full" eval policy, and creating a dataloader that can expand at runtime will help catch any potential issues that I may not be able to find by reviewing alone.

src/gepa/core/state.py
```diff
@@ -106,68 +119,127 @@ def load(run_dir: str) -> "GEPAState":
...
+            self.get_program_average_val_subset(program_idx)[0]
+            for program_idx in range(len(self.prog_candidate_val_subscores))
+        ]
+
+    def _update_pareto_front_for_val_id(
+        self,
+        val_id: DataId,
+        score: float,
+        program_idx: ProgramIdx,
+        outputs: dict[DataId, RolloutOutput] | None,
```
LakshyAAAgrawal: Given that this function is defined only for one val_id, the outputs should be just output: RolloutOutput?

src/gepa/core/state.py
```diff
@@ -106,68 +119,127 @@ def load(run_dir: str) -> "GEPAState":
...
+            self._update_pareto_front_for_val_id(val_id, score, new_program_idx, valset_outputs, run_dir, self.i + 1)
+        return new_program_idx
 
-                if self.best_outputs_valset is not None:
-                    self.best_outputs_valset[task_idx] = [(new_program_idx, valset_outputs[task_idx])]
 
-                if run_dir is not None:
-                    os.makedirs(os.path.join(run_dir, "generated_best_outputs_valset", f"task_{task_idx}"), exist_ok=True)
-                    with open(os.path.join(run_dir, "generated_best_outputs_valset", f"task_{task_idx}", f"iter_{self.i+1}_prog_{new_program_idx}.json"), "w") as f:
-                        json.dump(valset_outputs[task_idx], f, indent=4, default=json_default)
```
LakshyAAAgrawal, line 160: This is the place where the output was earlier being written.

src/gepa/strategies/batch_sampler.py
```diff
@@ -38,12 +48,12 @@ def _update_shuffled(self, trainset_size: int):
                 self.shuffled_ids.append(selected_id)
                 self.id_freqs[selected_id] += 1
 
-    def next_minibatch_indices(self, trainset_size: int, iteration: int) -> list[int]:
-        base_idx = iteration * self.minibatch_size
+    def next_minibatch_ids(self, loader: DataLoader[DataId, DataInst], state: GEPAState) -> list[DataId]:
```
LakshyAAAgrawal, line 58: The logic of both _update_shuffled and next_minibatch_ids assumes that the length of the trainset is fixed. I don't think this will work well when the trainset expands (the fix should be simple, just need to update shuffled ids every time next minibatch_ids is called, so that it continues to reflect the correct training dataset size).

src/gepa/proposer/reflective_mutation/base.py
```diff
@@ -6,11 +6,15 @@
...
 from gepa.core.adapter import Trajectory
 from gepa.core.state import GEPAState
+from gepa.strategies.eval_policy import EvaluationPolicy
 
 
 class CandidateSelector(Protocol):
     def select_candidate_idx(self, state: GEPAState) -> int: ...
 
+    def supports_eval_policy(eval_policy: EvaluationPolicy) -> bool:
+        return True
```
LakshyAAAgrawal: This shouldn't have a default implementation imo

src/gepa/core/engine.py
aria42: This eval policy actually is a sub-protocol for the `BatchSampler` and shares the method name, but I actually think it should just be its own protocol and agree name is confusing. The train sampling vs validation selection method have the same signatures, but semantics are different enough.

aria42:
> I think since the goal of the PR is to introduce dynamic valset, it would be useful to have at least one example run that actually uses a dynamic valset. 

Sounds good @LakshyAAAgrawal I'll add this and address other comments.

src/gepa/proposer/reflective_mutation/base.py
aria42: Removed in e0b75abe

src/gepa/strategies/batch_sampler.py
aria42: Resolved in 2f825d627c

src/gepa/core/state.py
aria42: I think this is still present it's just been moved into the function handling the Pareto update for a validation example. I don't believe the behavior here should'd ve changed.  

https://github.com/gepa-ai/gepa/pull/100/files#diff-b7b8ba47794881ef8b5fbe1e3194082101618e758b9b5781435a9c0d3f43dc74R189-R193

src/gepa/core/state.py
aria42: Resolved in 7b0243685

src/gepa/core/engine.py
aria42: Resolved in  e0b75abec and renamed to `get_eval_batch`

src/gepa/api.py
aria42: Resolved in c4e574d8

src/gepa/api.py
aria42: Resolved in 550a6399 which takes this suggestion

src/gepa/gepa_utils.py
aria42: Resolved in 7604b32 and added a test for this condition

aria42:
Hi @LakshyAAAgrawal, thanks for the thoughtful comments. I've addressed them individual commits, except for outputting the rollout for new programs, since I think that code is still present (just moved).  

For the larger request of adding a full e2e test that uses the dynamic data loading and the non-full validation policy. I've attempted this in  40d074141d3. It takes the RAG e2e test currently there and tests using dynamic data loading and a round-robin subset sample for validation.

This does exercise the code-paths, but it's obviously still an e2e "toy" example. Would love to flesh out something that works on an actual dataset, but this would require more work. I think building a "ELO"-style validation policy designed to estimate candidate ranking efficiently (e.g., something like TrueSkill)  and show you match performance with far fewer rollouts. But wondering if we should do that separately?

src/gepa/api.py
```diff
@@ -20,13 +23,19 @@
...
     RoundRobinReflectionComponentSelector,
 )
+from gepa.strategies.eval_policy import EvaluationPolicy, FullEvaluationPolicy, RoundRobinSampleEvaluationPolicy
 from gepa.utils import FileStopper, StopperProtocol
 
 
+class _ValEvaluationPolicyName(str, Enum):
+    FULL_EVAL = "full_eval"
+    ROUND_ROBIN_SAMPLE = "round_robin_sample"
+
```
LakshyAAAgrawal: One big issue with ROUND_ROBIN_SAMPLE is that it is probably not practically useful for real usecases, since it prioritizes evaluation on datapoints that have not been evaluated before, thereby providing zero support across different candidates.

However, it is certainly useful to test this functionality. I think it would be better to remove this from the enum, move the class itself to the place where this is being tested.

LakshyAAAgrawal:
@aria42 I am really sorry for the many back-forths. Since this PR is so big and touches almost every single file in the repository, I am trying to be extra critical, as reverting any of these decisions is going to be very difficult. I think we are very close to it, but just 1-2 more rounds.

I saw some major merge-conflicts after I merged some other PRs today, but I have pushed a fix for them already. Starting another review now.

src/gepa/api.py
```diff
@@ -20,13 +23,19 @@
     AllReflectionComponentSelector,
     RoundRobinReflectionComponentSelector,
 )
+from gepa.strategies.eval_policy import EvaluationPolicy, FullEvaluationPolicy, RoundRobinSampleEvaluationPolicy
 from gepa.utils import FileStopper, StopperProtocol
 
 
+class _ValEvaluationPolicyName(str, Enum):
```
LakshyAAAgrawal: Having the enum is nice, but considering that all the other options currently just accept "SpecialDType | str | None", let's remove the enum for consistency for now, and introduce a nicer configuration interface in another PR. See #3.

src/gepa/api.py
```diff
@@ -59,6 +69,8 @@ def optimize(
     # Reproducibility
     seed: int = 0,
     raise_on_exception: bool = True,
+    val_evaluation_policy: EvaluationPolicy[DataId, DataInst] | str | None = None,
+    val_evaluation_sample_size: int | None = None,
```
LakshyAAAgrawal: Once we remove round_robin from the main enum and as a main class in the repo, we can remove the val_evaluation_sample_size argument too.

LakshyAAAgrawal:
Dear @aria42,

I took a pass at changing/reverting several things as I reviewed the code. I request you to please review the changes I made in the series of commits: [[9028341](https://github.com/gepa-ai/gepa/pull/100/commits/9028341038e6cf53227c126f0ed240d7bbd5a62b), [2ab5e59](https://github.com/gepa-ai/gepa/pull/100/commits/2ab5e59f803fccdd5f6ed5bbed7478be423cd522), [04d24ef](https://github.com/gepa-ai/gepa/pull/100/commits/04d24ef2559f3f667d3344f6aa5fac6fc25dc91c)]. What are your thoughts about the reverts?

There are 2 newly added tests that are now failing due to my reversals, and I am wondering what are your thoughts about how to handle it?

aria42:
hey @LakshyAAAgrawal, the commits you added look good. Agree with encapsulating the scoring in the evaluation policy. Also on moving the richer example to tests since I agree it's mostly there to validate everything works. I think making something practically useful is more significant. 

The tests that failed after your changes were obviated due to your changes and aren't valid so I just removed them and unit tests pass.

tests/test_rag_adapter/test_rag_end_to_end.py
```diff
@@ -167,6 +187,89 @@ def get_collection_info(self) -> dict[str, Any]:
...
+        return "Gradient descent optimizes model parameters by minimizing the cost function iteratively."
+    return "This is a general AI/ML answer based on the provided context."
+
+
+def simple_reflection_lm(prompt):
+    """Simple reflection that suggests a better prompt."""
+    return json.dumps({"answer_generation": PREFERRED_DYNAMIC_PROMPT})
+
+
+class RAGTestAdapater(GenericRAGAdapter):
```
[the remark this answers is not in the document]
LakshyAAAgrawal: The corresponding change needs to be made at all places RAGTestAdapter is being used too, right?

src/gepa/proposer/merge.py
```diff
@@ -279,7 +295,10 @@ def propose(self, state: GEPAState) -> CandidateProposal | None:
             state.prog_candidate_val_subscores[id1],
             state.prog_candidate_val_subscores[id2],
         )
-        mini_devset = [self.valset[k] for k in subsample_ids]
+        mini_devset = self.valset.fetch(subsample_ids)
+        # below is a post condition of `select_eval_subsample_for_merged_program`
+        assert set(subsample_ids).issubset(state.prog_candidate_val_subscores[id1].keys())
+        assert set(subsample_ids).issubset(state.prog_candidate_val_subscores[id2])
```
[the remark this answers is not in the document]
LakshyAAAgrawal: @aria42 could you confirm this change is valid?

src/gepa/proposer/merge.py
```diff
@@ -214,37 +221,37 @@ def schedule_if_needed(self):
...
 
         remaining = num_subsample_ids - len(selected)
-        unused = list(all_indices - set(selected))
         if remaining > 0:
+            unused = [idx for idx in common_ids if idx not in selected]
             if len(unused) >= remaining:
                 selected += self.rng.sample(unused, k=remaining)
             else:
-                selected += self.rng.choices(list(all_indices), k=remaining)
+                selected += self.rng.choices(common_ids, k=remaining)
```
fgn, line 252: Consider what happens here if users passed `merge_val_overlap_floor=0`

See 
```python
        def has_val_support_overlap(id1: ProgramIdx, id2: ProgramIdx) -> bool:
            if self.val_overlap_floor <= 0:
                return True
            common_ids = set(state.prog_candidate_val_subscores[id1].keys()) & set(
                state.prog_candidate_val_subscores[id2].keys()
            )
            return len(common_ids) >= self.val_overlap_floor
```

Setting `merge_val_overlap_floor = 0` means "no minimum overlap requirement," but you still implicitly need at least one common validation item between the two parents to avoid an error here.

If `common_ids` is empty this will raise an IndexError.

LakshyAAAgrawal (dismissed):
Dear @aria42

Thank you so much for this wonderful, and incredibly detailed PR that takes GEPA towards having a very flexible API.

I am ready to merge. Could you please just address the 2-3 remaining unresolved conversations (Thanks a lot @fgn for the review)!

src/gepa/api.py
```diff
@@ -147,6 +155,10 @@ def optimize(
             "Since an adapter is provided, GEPA does not require a task LM to be provided. Please set the `task_lm` parameter to None."
         )
 
+    # cast data to DataLoader
+    trainset = ensure_loader(trainset)
+    valset = ensure_loader(valset or trainset)
```
fgn: This changes the previous behavior. If valsets `__len__` returns zero then it's falsy, just like `valset==[]` is falsy.
Before only when `valset is None` was is replaced with trainset. 

I could see this happening with the StagedDataloader. Nothing prevents it from having an initial size zero and grow?

tests/test_rag_adapter/test_rag_end_to_end.py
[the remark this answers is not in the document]
aria42: Yes correct, the commit it suggested looks like 8c5fbab does this correctly.

src/gepa/proposer/merge.py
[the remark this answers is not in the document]
aria42: It actually works without this fix since .issubset will pull the keys of the dict, but being explicit like in the commit is better!

```
>>> a = set("a")
>>> a.issubset({"a":1})
True
```

src/gepa/api.py
aria42: Good catch! Changed condition to only fallback to transit when valset is None to preserve existing semantics in  78f3b8ed5

src/gepa/proposer/merge.py
aria42: Good catch! I'm validating `val_overlap_floor` in the ctor to be positive which avoids the confusing semantics in 098df7347

aria42:
Thanks @LakshyAAAgrawal  and @fgn 

I've confirmed and resolved the open items from the copilot review. I also pushed up fixes for @fgn's comments and assuming those look good, I think everything open is resolved.

LakshyAAAgrawal:
Thank you so so much @aria42 for the PR, and @fgn for the reviews!

This is a large step towards making GEPA a very generalizable framework for all kinds of data-aware candidate and data selection straegies (including the rich literature on genetic algorithms). I am very excited about the next steps!

PR #109, opened 2025-10-13
epsilon greedy candidate selector

vacmar01:
I implemented a epsilon greedy candidate selector strategy, added it to the API and added some tests (for all candidate selectors). 

It is not fully documented yet, but I wanted to make sure that I'm on the right path with the API. I changed it a bit so that `optimize` also accepts a `CandidateSelector` subclass directly, so that we can specify further hyperparameters (like the epsilon in this case). 

What do you think, @LakshyAAAgrawal ? :)

Plan is to add more candidate selection strategies (Softmax, Thompson  Sampling, UCB1 etc.) to the repo and compare them to each other.

### DIFF
--- a/src/gepa/api.py
+++ b/src/gepa/api.py
@@ -223,17 +223,19 @@ def optimize(
     rng = random.Random(seed)
 
     if isinstance(candidate_selection_strategy, str):
-        candidate_selector_cls = {
-            "pareto": ParetoCandidateSelector(rng=rng),
-            "current_best": CurrentBestCandidateSelector(),
-            "epsilon_greedy": EpsilonGreedyCandidateSelector(epsilon=0.1, rng=rng),
-        }.get(candidate_selection_strategy)
+        factories = {
+            "pareto": lambda: ParetoCandidateSelector(rng=rng),
+            "current_best": lambda: CurrentBestCandidateSelector(),
+            "epsilon_greedy": lambda: EpsilonGreedyCandidateSelector(epsilon=0.1, rng=rng),
+        }
 
-        assert candidate_selector_cls is not None, (
-            f"Unknown candidate_selector strategy: {candidate_selection_strategy}. Supported strategies: 'pareto', 'current_best', 'epsilon_greedy'"
-        )
-
-        candidate_selector = candidate_selector_cls
+        try:
+            candidate_selector = factories[candidate_selection_strategy]()
+        except KeyError:
+            raise ValueError(
+                f"Unknown candidate_selector strategy: {candidate_selection_strategy}. "
+                "Supported strategies: 'pareto', 'current_best', 'epsilon_greedy'"
+            )
 
     if isinstance(candidate_selection_strategy, CandidateSelector):
         candidate_selector = candidate_selection_strategy


fgn:
Consider restricting ε-greedy exploration to a more refined pool of candidates. Currently, exploration may select any candidate, including those that are clearly suboptimal.
 
You might parameterize `EpsilonGreedyCandidateSelector` with a `pool` argument accepting values like `"pareto"`, `"top_k"`, or `"all"`, along with an optional `k` parameter for the `"top_k"` setting.

One suggestion could be to start with the Pareto set to ensure a good balance between exploration and exploitation, and then experiment with the `top_k` setting to fine-tune performance.

src/gepa/api.py
```diff
@@ -202,9 +206,22 @@ def optimize(
         valset = trainset
 
     rng = random.Random(seed)
-    candidate_selector = (
-        ParetoCandidateSelector(rng=rng) if candidate_selection_strategy == "pareto" else CurrentBestCandidateSelector()
-    )
+
+    if isinstance(candidate_selection_strategy, str):
+        candidate_selector_cls = {
```
fgn: Don't call it `_cls` here when it's not a class.

fgn: This will instantiate all three selectors.

Here is a lazy pattern that will instantiate at most one. 
```
factories = {
    "pareto": lambda: ParetoCandidateSelector(rng=rng),
    "best": lambda: CurrentBestCandidateSelector(),
    "epsilon_greedy": lambda: EpsilonGreedyCandidateSelector(epsilon=0.1, rng=rng),
}

try:
    candidate_selector = factories[candidate_selection_strategy]()
except KeyError:
    raise ValueError(
        f"Unknown candidate_selector strategy: {candidate_selection_strategy}. "
        "Supported strategies: 'pareto', 'best', 'epsilon_greedy'"
    )
```

vacmar01: Great review, thanks :) Will incorporate the changes.

LakshyAAAgrawal:
@vacmar01, This is great and agree with this direction. Thanks a lot @fgn for your review too!

@vacmar01, there's a large-ish refactor going on in #100, which will have a few changes to the API (specifically, you currently use per_program_tracked_scores from gepa_state, which is being removed). Could you refactor this around #100, which I plan to merge first, and then merge this PR next.

LakshyAAAgrawal:
Dear @vacmar01,

I have merged #100, and then merged the updated main back into this branch, trying to fix the merge conflicts as well. Could you please review if the changes I made align with your intented changes?

src/gepa/strategies/candidate_selector.py
```diff
@@ -31,3 +31,20 @@ def __init__(self):
...
+        self.epsilon = epsilon
+        if rng is None:
+            self.rng = random.Random(0)
+        else:
+            self.rng = rng
+
+    def select_candidate_idx(self, state: GEPAState) -> int:
+        assert len(state.program_full_scores_val_set) == len(state.program_candidates)
+        if self.rng.random() < self.epsilon:
+            return self.rng.randint(0, len(state.program_candidates) - 1)
```
LakshyAAAgrawal, line 48: Currently, the epsilon action has a uniform distribution over all program candidates. There can be several choices here itself:
1. Softmax of scores (with the best one scored -inf so that it is not selected here), that prioritizes the others in descending order of scores.
2. Using pareto-filter to remove all pareto-unoptimal and uniform distribution over them.
3. Actually using the pareto-based selector only in the epsilon condition.

It could be made fully configurable? I don't have a good intuition of what would be the best, but this is perhaps worth exploring. Happy to merge it without changes though.

vacmar01:
Hey @LakshyAAAgrawal, 

Im currently a bit under the weather, but I will take a look hopefully tomorrow. Hope that is still timely. 

Marius

vacmar01:
I just added the pattern for instantiation of the `CandidateSelector` that @fgn proposed above. I'm happy with the implementation as is. 

I also thought about expanding the curent EpsilonGreedy class with something like softmax sampling for the "non-best" candidates, but I think that is something for a future PR, right? Next up, I would like to evaluate the EpsilonGreedy strategy on LangProBe.

LakshyAAAgrawal:
@vacmar01, this looks ready to merge, for me. Kindly let me know if you are still making any changes, or shall I go ahead and merge!

vacmar01:
I'm happy. Can be merged from my side

PR #118, opened 2025-10-20
Reuse MLflow active run if exist

TomeHirata:
Currently, the experiment tracker always creates a new MLflow run, but it's more useful if users can log metrics to an existing active run to consolidate gepa metrics with other metrics.

### DIFF
--- a/src/gepa/logging/experiment_tracker.py
+++ b/src/gepa/logging/experiment_tracker.py
@@ -37,6 +37,8 @@ class ExperimentTracker:
         self.mlflow_tracking_uri = mlflow_tracking_uri
         self.mlflow_experiment_name = mlflow_experiment_name
 
+        self._created_mlflow_run = False
+
     def initialize(self):
         """Initialize the logging backends."""
         if self.use_wandb:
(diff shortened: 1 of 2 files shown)


LakshyAAAgrawal:
Hi @TomeHirata,

Wandb can log to some directories and continue runs from the same directory as well. Does MLFlow have support for such a feature? If yes, we could pass run_dir/log_dir argument from gepa.optimize to mlflow, and it can detect if a run already exists there or not. But run_dir/log_dir is not a mandatory argument, so this should work even without it (as it already does according to the PR).

TomeHirata:
In MLflow, a run is usually an immutable object. So even though there's a way to log additional logs to a completed run (similar to wandb), that is not what I meant to enable in this PR. The proposed change here is that when users have an active run before calling `gepa.optimize` (see the example below), we should log metrics to the existing active run rather than creating a nested one.

```python
with mlflow.start_run() as run: # mlflow run is created before calling gepa.optimize
  log_other_metrics(run)
  gepa.optimize(use_mlflow=True) # metrics are logged to the existing run
```

PR #119, opened 2025-10-20
Allowing custom batch_sampler

### LINKED ISSUE #113: Gepa optimize should accept a custom batch_sampler
I want to add a stratified batch_sampler. I can code it but gepa.optimize does not accept this a s a parameter.

I can do a PR if you think it makes sense.

Thanks,

CrisRodriguez:
Allowing custom batch_sampler
Closes #113

### DIFF
--- a/src/gepa/api.py
+++ b/src/gepa/api.py
@@ -13,9 +13,9 @@ from gepa.core.result import GEPAResult
 from gepa.logging.experiment_tracker import create_experiment_tracker
 from gepa.logging.logger import LoggerProtocol, StdOutLogger
 from gepa.proposer.merge import MergeProposer
-from gepa.proposer.reflective_mutation.base import BatchSampler, LanguageModel, ReflectionComponentSelector
+from gepa.proposer.reflective_mutation.base import LanguageModel, ReflectionComponentSelector
 from gepa.proposer.reflective_mutation.reflective_mutation import ReflectiveMutationProposer
-from gepa.strategies.batch_sampler import EpochShuffledBatchSampler
+from gepa.strategies.batch_sampler import BatchSampler, EpochShuffledBatchSampler
 from gepa.strategies.candidate_selector import CurrentBestCandidateSelector, ParetoCandidateSelector
 from gepa.strategies.component_selector import (
     AllReflectionComponentSelector,


src/gepa/api.py
```diff
@@ -220,7 +222,13 @@ def optimize(
 
         module_selector = module_selector_cls()
 
-    batch_sampler = EpochShuffledBatchSampler(minibatch_size=reflection_minibatch_size, rng=rng)
+    if batch_sampler is None:
+        batch_sampler = EpochShuffledBatchSampler(minibatch_size=reflection_minibatch_size, rng=rng)
+    else:
+        logger.log(
+            "WARNING: Both 'batch_sampler' and 'reflection_minibatch_size' were provided. "
+            "Ignoring 'reflection_minibatch_size' since 'batch_sampler' takes precedence."
```
LakshyAAAgrawal: Let's throw an error instead of warning. Make the corresponding change in the documentation above. I prefer to prevent the user from executing something the developer knows is going to lead to unexpected outcomes.

CrisRodriguez: @LakshyAAAgrawal 
Done, thanks

src/gepa/api.py
```diff
@@ -33,7 +33,8 @@ def optimize(
     reflection_lm: LanguageModel | str | None = None,
     candidate_selection_strategy: str = "pareto",
     skip_perfect_score=True,
-    reflection_minibatch_size=3,
+    batch_sampler: BatchSampler | None = None,
+    reflection_minibatch_size=None,
```
LakshyAAAgrawal: Could you retain the reflection_minibatch_size argument? We shouldn't delete and break backwards compatibility.

Following the other examples like "candidate_selection_strategy", can you make this also accept `BatchSampler | str | None`, where it has a default string argument that matches "epoch_shuffled"

src/gepa/api.py
```diff
@@ -220,7 +222,13 @@ def optimize(
 
         module_selector = module_selector_cls()
 
-    batch_sampler = EpochShuffledBatchSampler(minibatch_size=reflection_minibatch_size, rng=rng)
+    if batch_sampler and reflection_minibatch_size:
```
LakshyAAAgrawal: This should become
```
if batch_sampler == 'epoch_shuffled':
    reflection_minibatch_size = reflection_minibatch_size or 3
else:
   assert reflection_minibatch_size is None
```

src/gepa/api.py
CrisRodriguez: Hi,
Thanks for the review.


> Could you retain the reflection_minibatch_size argument? We shouldn't delete and break backwards compatibility.
I did not delete it, looks like a git diff vis pb.

> Following the other examples like "candidate_selection_strategy", can you make this also accept BatchSampler | str | None, where it has a default string argument that matches "epoch_shuffled"

Sure. Then it does not make sense to keep the None. Implementing as `batch_sampler: BatchSampler | str = "epoch_shuffled"`

LakshyAAAgrawal:
@CrisRodriguez, thank you so much for your changes! I will be merging #100 first, could you please plan to rebase this over it as I expect there to be merge conflicts.

LakshyAAAgrawal:
Dear @CrisRodriguez, thank you so much for your kind contribution!

PR #120, opened 2025-10-21
Add MLflow prompt optimization using GEPA on README

TomeHirata:
In MLflow 3.5.0, we released a new API `mlflow.genai.optimize_prompts` that enable MLflow users to optimize their prompts using the GEPA algorithm ([doc](https://mlflow.org/docs/latest/genai/prompt-registry/optimize-prompts/)). This PR adds the link and explanation in the "GEPA Integrations" section.


LakshyAAAgrawal:
Could you consider adding the adapter to src/gepa/adapters?

PR #131, opened 2025-11-08
Add logging option for individual valset scores

LakshyAAAgrawal:
Added a flag to log individual validation set scores and programs.


LakshyAAAgrawal:
CC @Shangyint

PR #147, opened 2025-12-19
feat(default_adapter): configurable eval returns score + feedback (#62)

### LINKED ISSUE #62: DefaultAdapter's eval function should be configurable, it can return score and feedback

villurignanesh:
Fixes #62

Hi @LakshyAAAgrawal, could you please take a look?

**What changed**
- Added a configurable per-example evaluation hook to DefaultAdapter so scoring logic is user-defined.
- Updated evaluate() to return per-example score and feedback in the rollout output/trace to support richer reflection.


LakshyAAAgrawal:
Thanks a lot for the PR @villurignanesh!

PR #155, opened 2026-01-04
Add callback system for optimization instrumentation

dbreunig:
Seeing reflections and evaluations in the CLI is nice, but we can enable enhanced logging and other visualizations with callback functionality. 

I like this feature because it opens up the potential for people to better display *what* GEPA is doing under the hood.

- Add GEPACallback protocol with 20 callback methods covering:
  - Optimization lifecycle (start/end)
  - Iteration lifecycle (start/end)
  - Candidate selection and minibatch sampling
  - Evaluation events (start/end)
  - Reflection events (dataset built, proposal start/end)
  - Acceptance/rejection events
  - Merge events (attempted/accepted/rejected)
  - State events (Pareto front updated, state saved)
  - Budget tracking and error handling

- Add CompositeCallback helper class and notify_callbacks utility
- Add callbacks parameter to optimize() function
- Integrate callback notifications throughout:
  - GEPAEngine.run() - main optimization loop
  - ReflectiveMutationProposer.propose() - mutation workflow
  - MergeProposer.propose() - merge workflow

- Add 34 comprehensive tests covering all callback functionality

Here is [a demo implementation](https://gist.github.com/dbreunig/cca951889cb595bc57a412477742d3a4), using the callback functionality to log to file. This is [the log file it created](https://gist.github.com/dbreunig/19c02310a71f457922831169b580c6a6).

### DIFF
--- a/src/gepa/core/engine.py
+++ b/src/gepa/core/engine.py
@@ -328,6 +328,26 @@ class GEPAEngine(Generic[DataId, DataInst, Trajectory, RolloutOutput]):
             ),
         )
 
+        # Notify callbacks of seed candidate's initial valset evaluation (iteration 0)
+        # This provides the baseline performance before any optimization
+        seed_scores = state.prog_candidate_val_subscores[0]
+        notify_callbacks(
+            self.callbacks,
+            "on_valset_evaluated",
+            ValsetEvaluatedEvent(
+                iteration=0,
+                candidate_idx=0,
+                candidate=self.seed_candidate,
+                scores_by_val_id=dict(seed_scores),
+                average_score=base_val_avg,
+                num_examples_evaluated=len(seed_scores),
+                total_valset_size=len(valset),
+                parent_ids=[],
+                is_best_program=True,  # Seed is always best at iteration 0
+                outputs_by_val_id=None,  # Outputs not tracked at initialization unless track_best_outputs=True
+            ),
+        )
+
         # Register budget hook to fire on_budget_updated callback in real-time
         def budget_hook(new_total: int, delta: int) -> None:
             notify_callbacks(
--- a/tests/test_callbacks_integration/test_callbacks_with_optimization.py
+++ b/tests/test_callbacks_integration/test_callbacks_with_optimization.py
@@ -1,0 +1,368 @@
+# Copyright (c) 2025 Lakshya A Agrawal and the GEPA contributors
+# https://github.com/gepa-ai/gepa
+
+"""Integration tests for callbacks during a real optimization run.
+
+These tests verify that callbacks are properly invoked during the GEPA
+optimization process with real data and LLM calls (mocked for determinism).
+
+Note: This test reuses the LLM cache from test_aime_prompt_optimization to
+avoid duplicating the cache data.
+"""
+
+from pathlib import Path
+
+import pytest
+
+# Reuse the AIME test's LLM cache directory
+RECORDER_DIR = Path(__file__).parent.parent / "test_aime_prompt_optimization"
+
+
+class RecordingCallback:
+    """A callback that records all method calls for integration testing."""
+
+    def __init__(self):
+        self.calls = []
+
+    def _record(self, method_name, event):
+        self.calls.append((method_name, dict(event)))
+
+    def get_calls(self, method_name):
+        """Get all calls to a specific method."""
+        return [kwargs for name, kwargs in self.calls if name == method_name]
+
+    def on_optimization_start(self, event):
+        self._record("on_optimization_start", event)
+
+    def on_optimization_end(self, event):
+        self._record("on_optimization_end", event)
+
+    def on_iteration_start(self, event):
+        self._record("on_iteration_start", event)
+
+    def on_iteration_end(self, event):
+        self._record("on_iteration_end", event)
+
+    def on_candidate_selected(self, event):
+        self._record("on_candidate_selected", event)
+
+    def on_minibatch_sampled(self, event):
+        self._record("on_minibatch_sampled", event)
+
+    def on_evaluation_start(self, event):
+        self._record("on_evaluation_start", event)
+
+    def on_evaluation_end(self, event):
+        self._record("on_evaluation_end", event)
+
+    def on_evaluation_skipped(self, event):
+        self._record("on_evaluation_skipped", event)
+
+    def on_reflective_dataset_built(self, event):
+        self._record("on_reflective_dataset_built", event)
+
+    def on_proposal_start(self, event):
+        self._record("on_proposal_start", event)
+
+    def on_proposal_end(self, event):
+        self._record("on_proposal_end", event)
+
+    def on_candidate_accepted(self, event):
+        self._record("on_candidate_accepted", event)
+
+    def on_candidate_rejected(self, event):
+        self._record("on_candidate_rejected", event)
+
+    def on_merge_attempted(self, event):
+        self._record("on_merge_attempted", event)
+
+    def on_merge_accepted(self, event):
+        self._record("on_merge_accepted", event)
+
+    def on_merge_rejected(self, event):
+        self._record("on_merge_rejected", event)
+
+    def on_pareto_front_updated(self, event):
+        self._record("on_pareto_front_updated", event)
+
+    def on_state_saved(self, event):
+        self._record("on_state_saved", event)
+
+    def on_budget_updated(self, event):
+        self._record("on_budget_updated", event)
+
+    def on_error(self, event):
+        self._record("on_error", event)
+
+    def on_valset_evaluated(self, event):
+        self._record("on_valset_evaluated", event)
+
+
+@pytest.fixture(scope="module")
+def recorder_dir() -> Path:
+    """Provides the path to the recording directory and ensures it exists."""
+    RECORDER_DIR.mkdir(parents=True, exist_ok=True)
+    return RECORDER_DIR
+
+
+def test_callbacks_during_optimization(mocked_lms, recorder_dir):
+    """
+    Tests that callbacks are properly invoked during GEPA optimization.
+
+    This test verifies:
+    1. on_optimization_start is called once at the beginning
+    2. on_optimization_end is called once at the end
+    3. on_iteration_start/end are called for each iteration
+    4. Evaluation callbacks are called during candidate evaluation
+    5. Budget callbacks are called when metric calls are made
+    6. Pareto front update callbacks are called when new candidates are accepted
+    """
+    import gepa
+    from gepa.adapters.default_adapter.default_adapter import DefaultAdapter
+
+    # 1. Setup: Unpack fixtures and load data
+    task_lm, reflection_lm = mocked_lms
+    adapter = DefaultAdapter(model=task_lm)
+
+    print("Initializing AIME dataset...")
+    trainset, valset, _ = gepa.examples.aime.init_dataset()
+    trainset = trainset[:10]
+    valset = valset[:10]
+
+    seed_prompt = {
+        "system_prompt": "You are a helpful assistant. You are given a question and you need to answer it. The answer should be given at the end of your response in exactly the format '### <final answer>'"
+    }
+
+    # Create callback to record all events
+    callback = RecordingCallback()
+
+    # 2. Execution: Run the core optimization logic with callbacks
+    print("Running GEPA optimization process with callbacks...")
+    gepa_result = gepa.optimize(
+        seed_candidate=seed_prompt,
+        trainset=trainset,
+        valset=valset,
+        adapter=adapter,
+        max_metric_calls=30,
+        reflection_lm=reflection_lm,
+        display_progress_bar=True,
+        callbacks=[callback],
+    )
+
+    # 3. Assertions: Verify callbacks were invoked correctly
+
+    # Optimization lifecycle callbacks
+    opt_start_calls = callback.get_calls("on_optimization_start")
+    assert len(opt_start_calls) == 1, "on_optimization_start should be called exactly once"
+    assert opt_start_calls[0]["trainset_size"] == 10
+    assert opt_start_calls[0]["valset_size"] == 10
+    assert opt_start_calls[0]["seed_candidate"] == seed_prompt
+
+    opt_end_calls = callback.get_calls("on_optimization_end")
+    assert len(opt_end_calls) == 1, "on_optimization_end should be called exactly once"
+    assert "best_candidate_idx" in opt_end_calls[0]
+    assert "total_iterations" in opt_end_calls[0]
+    assert "total_metric_calls" in opt_end_calls[0]
+    assert opt_end_calls[0]["total_metric_calls"] > 0
+
+    # Iteration callbacks
+    iter_start_calls = callback.get_calls("on_iteration_start")
+    iter_end_calls = callback.get_calls("on_iteration_end")
+    assert len(iter_start_calls) >= 1, "at least one iteration should have started"
+    assert len(iter_start_calls) == len(iter_end_calls), "each iteration_start should have a matching iteration_end"
+
+    # Verify iterations are numbered correctly (1-indexed)
+    for i, call in enumerate(iter_start_calls):
+        assert call["iteration"] == i + 1, f"iteration should be 1-indexed, got {call['iteration']} for index {i}"
+
+    # Candidate selection callbacks
+    candidate_selected_calls = callback.get_calls("on_candidate_selected")
+    assert len(candidate_selected_calls) >= 1, "at least one candidate should be selected"
+
+    # Minibatch sampling callbacks
+    minibatch_calls = callback.get_calls("on_minibatch_sampled")
+    assert len(minibatch_calls) >= 1, "at least one minibatch should be sampled"
+    for call in minibatch_calls:
+        assert call["trainset_size"] == 10
+
+    # Evaluation callbacks
+    eval_start_calls = callback.get_calls("on_evaluation_start")
+    eval_end_calls = callback.get_calls("on_evaluation_end")
+    assert len(eval_start_calls) >= 1, "at least one evaluation should start"
+    assert len(eval_end_calls) >= 1, "at least one evaluation should end"
+
+    # Budget callbacks
+    budget_calls = callback.get_calls("on_budget_updated")
+    assert len(budget_calls) >= 1, "budget should be updated at least once"
+    # Budget should increase monotonically
+    prev_used = 0
+    for call in budget_calls:
+        assert call["metric_calls_used"] >= prev_used, "budget should increase monotonically"
+        prev_used = call["metric_calls_used"]
+
+    # Pareto front callbacks
+    pareto_calls = callback.get_calls("on_pareto_front_updated")
+    assert len(pareto_calls) >= 1, "pareto front should be updated at least once"
+    for call in pareto_calls:
+        assert "new_front" in call
+        assert "displaced_candidates" in call
+
+    # Valset evaluated callbacks
+    valset_calls = callback.get_calls("on_valset_evaluated")
+    assert len(valset_calls) >= 1, "at least one valset evaluation should occur"
+
+    # Verify seed candidate valset evaluation is included (iteration 0)
+    seed_valset_calls = [c for c in valset_calls if c["iteration"] == 0]
+    assert len(seed_valset_calls) == 1, "seed candidate valset evaluation should be called at iteration 0"
+    seed_call = seed_valset_calls[0]
+    assert seed_call["candidate_idx"] == 0, "seed candidate should have index 0"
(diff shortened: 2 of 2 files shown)


src/gepa/core/callbacks.py
```diff
@@ -0,0 +1,529 @@
...
+        parent_idx: int,
+    ) -> None:
+        """Called when a new candidate is accepted.
+
+        Args:
+            iteration: Current iteration number.
+            new_candidate_idx: Index assigned to the new candidate.
+            new_score: The new candidate's validation score.
+            parent_idx: Index of the parent candidate.
+        """
```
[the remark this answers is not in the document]
dbreunig: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/gepa-ai/gepa/pull/155#discussion_r2660884769)

src/gepa/proposer/merge.py
```diff
@@ -339,7 +342,24 @@ def has_val_support_overlap(id1: ProgramIdx, id2: ProgramIdx) -> bool:
         id2_sub_scores = [state.prog_candidate_val_subscores[id2][k] for k in subsample_ids]
         state.full_program_trace[-1]["subsample_ids"] = subsample_ids
 
+        # Notify evaluation start for merged candidate
+        notify_callbacks(
+            self.callbacks,
+            "on_evaluation_start",
+            iteration=i,
+            candidate_idx=id1,  # Use first parent as reference
```
LakshyAAAgrawal: Is there a better way to handle candidate_idx here, since this is a different idx we are evaluating.

src/gepa/core/callbacks.py
```diff
@@ -0,0 +1,529 @@
...
+            parent_indices: Indices of the parent candidates.
+            reason: Explanation for rejection.
+        """
+        ...
+
+    # =========================================================================
+    # State Events
+    # =========================================================================
+
+    def on_pareto_front_updated(
```
LakshyAAAgrawal: This is defined but does not seem to be called anywhere. Perhaps it needs to be invoked within engine.py OR state.py

src/gepa/core/engine.py
```diff
@@ -309,18 +387,47 @@ def valset_evaluator(
...
                     new_program=proposal.candidate,
                     state=state,
                     parent_program_idx=proposal.parent_program_ids,
                 )
+                proposal_accepted = True
+
+                # Notify budget updated
+                notify_callbacks(
+                    self.callbacks,
+                    "on_budget_updated",
```
LakshyAAAgrawal: Budget update callback when candidates are accepted, whereas the budget has been consumed even when candidate has not been accepted.

LakshyAAAgrawal: It could be useful to track state.total_num_evals and track budget using that.

LakshyAAAgrawal:
There are situations where no trajectories are captured. We can consider adding something like `on_evaluation_skipped` event, for when evaluations are skipped due to missing trajectories or other reasons.

One such example is https://github.com/gepa-ai/gepa/blob/main/src/gepa/proposer/reflective_mutation/reflective_mutation.py#L113-L115

src/gepa/proposer/merge.py
dbreunig: Nice catch. I'd like to capture both parent ids, as well. Perhaps we add a parameter `parent_ids: list[int] | None`. That lets us path both in merge and a single parent in mutation.

src/gepa/core/callbacks.py
dbreunig: Am adding to engine.py so we don't have to pass new info to state. Is being called after `_run_full_eval_and_add()`, taking snapshots of the front before and after update so we can pass back displaced candidate info in the callback.

src/gepa/core/engine.py
dbreunig: I like the idea of tracking `state.total_num_evals`, even though it ends up adding some hooks to the state and registering the hooks in the engine.

An alternative would be passing callbacks to proposers, but that feels less clean.

I implemented the hooks approach to track `state.total_num_evals`. This requires an inelegant removal of the budget hooks from the state during `save` (the lambdas they contain can't be serialized). I've added test coverage for this, though.

Take a look.

dbreunig:
> There are situations where no trajectories are captured. We can consider adding something like `on_evaluation_skipped` event, for when evaluations are skipped due to missing trajectories or other reasons.
> 
> One such example is https://github.com/gepa-ai/gepa/blob/main/src/gepa/proposer/reflective_mutation/reflective_mutation.py#L113-L115

Yes, there's two places where this is relevant: when no trajectories are captured or when all subsample scores are perfect.

Implemented this.

src/gepa/core/engine.py
```diff
@@ -140,6 +150,24 @@ def _run_full_eval_and_add(
...
+        for program_set in front_after.values():
+            candidates_after.update(program_set)
+
+        new_front = sorted(candidates_after)
+        displaced_candidates = sorted(candidates_before - candidates_after)
+
+        notify_callbacks(
+            self.callbacks,
+            "on_pareto_front_updated",
+            iteration=state.i + 1,
```
LakshyAAAgrawal: Is the `state.i + 1` valid here?

src/gepa/core/engine.py
```diff
@@ -274,22 +345,57 @@ def valset_evaluator(
...
+
+                                # Notify merge accepted
+                                notify_callbacks(
+                                    self.callbacks,
+                                    "on_merge_accepted",
+                                    iteration=state.i,
+                                    new_candidate_idx=new_idx,
+                                    parent_indices=proposal.parent_program_ids,
+                                )
+                                notify_callbacks(
```
LakshyAAAgrawal, line 451: Since there are several arguments that a callback is called with, we can consider creating a dataclass / TypedDict or some structure, that will contain all the arguments, and then we call the callbacks with an instance of that structure. Allows to easily extend/modify the parameters passed to callbacks.

src/gepa/core/engine.py
```diff
@@ -274,22 +345,57 @@ def valset_evaluator(
...
+                                    self.callbacks,
+                                    "on_candidate_accepted",
+                                    iteration=state.i,
+                                    new_candidate_idx=new_idx,
+                                    new_score=new_sum,
+                                    parent_idx=proposal.parent_program_ids[0],
```
LakshyAAAgrawal: Since a candidate can have multiple parents, wherever parent_idx is a parameter, it should be a Sequence[ProgramIdx], not just one value.

src/gepa/core/engine.py
dbreunig: No. Bug. Fixing.

src/gepa/core/engine.py
dbreunig: Fixed

src/gepa/core/engine.py
dbreunig: There's not a lot of overlap between the function parameters, unfortunately, so it doesn't make sense to create one dataclass or even one dataclass per category. [Here's a matrix for what's being passed back, by function](https://docs.google.com/spreadsheets/d/1kPM2GuwWoYXeyu7bn0txaoD8yfpsGNGPz4ne-5R8MGc/edit?gid=631833450#gid=631833450).

dbreunig: I implemented TypedDicts for each method.

LakshyAAAgrawal: Thanks a lot for sharing the Matrix. This is very useful.

Do you think there's merit in passing the GEPAState in as many of these callbacks as possible? One of the benefits I had earlier imagined of having a callback system like this was a callback implementor to store things in the state, which can also be serialized into the directory. We can have some additional field in GEPAState like "user_state" which is just a keystore, and the callbacks can manage it.

We don't need to implement that right now, but passing in state in as many callbacks as possible could enable that.

What do you think?

dbreunig: Oh, that's an interesting idea. I can imagine a few applications there.

The question it raises is: can you mutate and pass back the state? I don't immediately have a use case for that, but it seems intuitively useful.

LakshyAAAgrawal: Since GEPAState is a stateful mutable object, which in this proposed future will have a field called "user_state" that will be a dict, it will be persistent across all callback calls, as well as across serialisation/deserialisation.

raveeshbhalla:
## Callback Gap Analysis for Dashboard/Observability Integration

Hi team! I'm Claude Code, working on integrating a real-time observability dashboard with GEPA using the new callback system in this PR. I wanted to share some gaps I discovered that make full dashboard integration challenging compared to the old observer pattern.

### Missing Callbacks

#### 1. `on_seed_validated` (or similar)

**Why it matters for observability:**
The seed score is the baseline against which all optimization progress is measured. Without it, users can't answer the fundamental question: "How much did optimization actually improve my prompt?" A dashboard needs to display the starting point so users can visualize the optimization trajectory—seeing scores climb from seed → iteration 1 → iteration 2 → final. This is especially critical for debugging: if the seed score is already high, users know there's limited room for improvement; if it's low, they can set expectations accordingly.

**What our fork does:** Fires `on_seed_validation(score, predictions)` when the seed candidate is evaluated, allowing the dashboard to show the baseline score prominently and include it in performance comparison charts.

**What isn't possible with PR #155:** No callback fires when the seed is validated. The dashboard shows "N/A" for seed score and can't render before/after comparisons. Users have no visibility into their starting point.

---

#### 2. `on_valset_evaluated` (or similar)

**Why it matters for observability:**
The validation set evaluation is the ground truth for whether optimization succeeded. Minibatch scores during training can be noisy or overfit to specific examples. The valset evaluation answers: "Does this optimized prompt actually generalize?" For a complete observability story, users need to see side-by-side: seed valset score vs. optimized valset score. This is the metric that matters for deciding whether to deploy the optimized prompt.

**What our fork does:** Fires `on_valset_eval(candidate_idx, scores, predictions)` for both seed and final candidates on the validation set, enabling a "Lineage" view that shows per-example performance comparison between seed and optimized prompts.

**What isn't possible with PR #155:** `on_evaluation_end` only fires for minibatch evaluations during optimization loops, not for the final valset comparison. The dashboard's "Lineage" tab shows "No evaluations to compare" because it filters for valset-phase evaluations that never arrive. Users can't see the final validation performance or compare seed vs. optimized on held-out data.

---

#### 3. `on_example_evaluated` with iteration/candidate context

**Why it matters for observability:**
Per-example visibility is crucial for debugging prompt optimization. When a proposed prompt scores lower than expected, users need to inspect individual examples: "Which inputs did it fail on? What did it predict? Why did it get that score?" This granular view helps users understand *why* a candidate was rejected and iterate on their metric or training data. Without per-example predictions tied to specific candidates, users only see aggregate scores with no ability to drill down.

**What our fork does:** Captures predictions inline during evaluation with full context (iteration, candidate_idx, example_inputs, prediction, score), allowing the dashboard to show a detailed breakdown of each candidate's performance with expandable rows showing input → prediction → score.

**What isn't possible with PR #155:** `on_evaluation_end` provides batch scores but no per-example predictions or inputs. The `on_lm_call` callback fires during evaluation but lacks `iteration` or `candidate_idx` context—with parallel execution, there's no way to reliably associate which LM output belongs to which evaluation run. The dashboard shows "No prediction captured" for all proposed candidates, leaving users blind to what the candidate actually produced.

**Suggestion:** Either:
- Add `iteration` and `candidate_idx` to `on_lm_call` when called during evaluation, OR
- Add a new `on_example_evaluated(iteration, candidate_idx, example_inputs, prediction, score)` callback

---

### Summary Table

| Capability | Our Fork (Observer) | PR #155 Callbacks |
|------------|---------------------|-------------------|
| Seed validation score | ✅ `on_seed_validation` | ❌ Missing |
| Valset evaluation | ✅ `on_valset_eval` | ❌ Missing |
| Per-example predictions during eval | ✅ Captured with full context | ⚠️ `on_lm_call` lacks iteration/candidate context |
| Minibatch evaluations | ✅ | ✅ `on_evaluation_end` |
| LM calls | ✅ | ✅ `on_lm_call` |

---

### Workaround Attempts

1. **Tried capturing LM outputs during evaluation** via `on_lm_call` and correlating with `on_evaluation_end`, but parallel execution means outputs arrive in arbitrary order with no way to match them to specific examples or candidates.

2. **Using reflective dataset** (`on_reflective_dataset_built`) to get parent candidate evaluations works well since it includes `{Inputs, Generated Outputs, Feedback}` per example. However, proposed candidate evaluations only receive aggregate scores without per-example data.

---

### Impact on User Experience

With PR #155 callbacks, an observability dashboard can show:
- ✅ Iteration-by-iteration minibatch scores (aggregate only)
- ✅ LM call traces (but not tied to specific evaluations)
- ✅ Candidate acceptance/rejection decisions
- ❌ **Seed baseline performance** — users can't see their starting point
- ❌ **Final optimized vs seed comparison** — users can't validate generalization
- ❌ **Per-example predictions for proposed candidates** — users can't debug why candidates fail

These gaps mean users lose visibility into the most important questions: "Where did I start?", "Did optimization actually help on held-out data?", and "Why did this candidate get rejected?"

---

Would love to discuss if adding these callbacks (or extending existing ones) would be feasible! Happy to help test any changes.

raveeshbhalla:
## Proposed Changes (Follow-up)

Based on my analysis of the PR diff, here are the specific changes that would address the gaps I identified:

---

### 1. Add `on_seed_validated` callback

**Location:** `src/gepa/core/state.py` in `initialize_gepa_state()` function

**Current code (around line 195-225):**
```python
eval_result = valset_evaluator(seed_candidate)
if run_dir is not None:
    write_eval_outputs_to_directory(
        eval_result.outputs_by_val_id, os.path.join(run_dir, "generated_best_outputs_valset")
    )

num_evals_run += len(eval_result.scores_by_val_id)
```

**Proposed change:**

First, add to `callbacks.py`:
```python
class SeedValidatedEvent(TypedDict):
    """Event for on_seed_validated callback."""
    seed_candidate: dict[str, str]
    scores: list[float]
    outputs: dict[Any, Any]  # outputs_by_val_id
    total_score: float
```

Then modify `initialize_gepa_state()` to accept and fire callback:
```python
def initialize_gepa_state(
    run_dir: str | None,
    logger: LoggerProtocol,
    seed_candidate: dict[str, str],
    valset_evaluator: Callable[...],
    track_best_outputs: bool = False,
    frontier_type: FrontierType = "instance",
    callbacks: list | None = None,  # NEW PARAMETER
) -> GEPAState[RolloutOutput, DataId]:
    # ... existing code ...
    
    eval_result = valset_evaluator(seed_candidate)
    
    # NEW: Notify seed validated
    notify_callbacks(
        callbacks,
        "on_seed_validated",
        SeedValidatedEvent(
            seed_candidate=seed_candidate,
            scores=list(eval_result.scores_by_val_id.values()),
            outputs=eval_result.outputs_by_val_id,
            total_score=sum(eval_result.scores_by_val_id.values()),
        ),
    )
    
    # ... rest of existing code ...
```

Also update `engine.py` to pass callbacks to `initialize_gepa_state()`.

---

### 2. Add `on_valset_evaluated` callback

**Location:** `src/gepa/core/engine.py` after optimization completes

**Current code (around end of `run()` method):**
```python
state.save(self.run_dir)

# Notify optimization end
best_candidate_idx = self.val_evaluation_policy.get_best_program(state)
notify_callbacks(
    self.callbacks,
    "on_optimization_end",
    OptimizationEndEvent(...),
)

return state
```

**Proposed change:**

First, add to `callbacks.py`:
```python
class ValsetEvaluatedEvent(TypedDict):
    """Event for on_valset_evaluated callback."""
    candidate_idx: int
    candidate: dict[str, str]
    phase: str  # "seed" or "final"
    scores_by_val_id: dict[Any, float]
    outputs_by_val_id: dict[Any, Any]
    total_score: float
    is_best: bool
```

Then add to `engine.py` before returning:
```python
state.save(self.run_dir)

# NEW: Fire valset evaluation for final comparison
best_candidate_idx = self.val_evaluation_policy.get_best_program(state)
best_candidate = state.full_program_bank[best_candidate_idx]
best_scores = state.prog_candidate_val_subscores[best_candidate_idx]

notify_callbacks(
    self.callbacks,
    "on_valset_evaluated",
    ValsetEvaluatedEvent(
        candidate_idx=best_candidate_idx,
        candidate=best_candidate,
        phase="final",
        scores_by_val_id=best_scores,
        outputs_by_val_id=state.prog_candidate_outputs.get(best_candidate_idx, {}),
        total_score=sum(best_scores.values()),
        is_best=True,
    ),
)

# Existing optimization_end callback
notify_callbacks(
    self.callbacks,
    "on_optimization_end",
    OptimizationEndEvent(...),
)
```

---

### 3. Add per-example context to `on_evaluation_end` (or new `on_example_evaluated`)

**Option A: Extend `EvaluationEndEvent`**

Add fields to the existing event in `callbacks.py`:
```python
class EvaluationEndEvent(TypedDict):
    """Event for on_evaluation_end callback."""
    iteration: int
    candidate_idx: int | None
    scores: list[float]
    has_trajectories: bool
    parent_ids: Sequence[ProgramIdx]
    # NEW FIELDS:
    example_ids: list[Any]  # The IDs of examples evaluated
    example_inputs: list[dict[str, Any]]  # The input data for each example
    predictions: list[Any] | None  # The model outputs, if available
```

**Location to modify:** `src/gepa/proposer/reflective_mutation/reflective_mutation.py`

Change evaluation callbacks to include per-example data:
```python
# After evaluating new candidate
eval_new = self.adapter.evaluate(minibatch, new_candidate, capture_traces=False)

notify_callbacks(
    self.callbacks,
    "on_evaluation_end",
    EvaluationEndEvent(
        iteration=i,
        candidate_idx=None,
        scores=eval_new.scores,
        has_trajectories=False,
        parent_ids=[curr_prog_id],
        # NEW:
        example_ids=subsample_ids,
        example_inputs=[dict(ex) for ex in minibatch],
        predictions=eval_new.outputs,  # Include the actual predictions
    ),
)
```

**Option B: New granular callback**

Add a new callback that fires per-example:
```python
class ExampleEvaluatedEvent(TypedDict):
    """Event for on_example_evaluated callback."""
    iteration: int
    candidate_idx: int | None
    example_id: Any
    example_inputs: dict[str, Any]
    prediction: Any
    score: float

# Fire for each example after evaluation
for idx, (ex_id, score) in enumerate(zip(subsample_ids, eval_new.scores)):
    notify_callbacks(
        self.callbacks,
        "on_example_evaluated",
        ExampleEvaluatedEvent(
            iteration=i,
            candidate_idx=None,
            example_id=ex_id,
            example_inputs=dict(minibatch[idx]),
            prediction=eval_new.outputs[idx],
            score=score,
        ),
    )
```

---

### Summary of Changes Required

| File | Change |
|------|--------|
| `src/gepa/core/callbacks.py` | Add `SeedValidatedEvent`, `ValsetEvaluatedEvent`, extend `EvaluationEndEvent` or add `ExampleEvaluatedEvent` |
| `src/gepa/core/state.py` | Add `callbacks` parameter to `initialize_gepa_state()`, fire `on_seed_validated` |
| `src/gepa/core/engine.py` | Pass callbacks to `initialize_gepa_state()`, fire `on_valset_evaluated` before optimization ends |
| `src/gepa/proposer/reflective_mutation/reflective_mutation.py` | Include `example_ids`, `example_inputs`, `predictions` in `on_evaluation_end` events |
| `src/gepa/proposer/merge.py` | Same changes for merge evaluations |

Happy to discuss or help test these changes!

LakshyAAAgrawal:
Thanks a lot for the detailed comments, @raveeshbhalla!

> 1. on_seed_validated (or similar)

GEPA will soon allow users to provide multiple seed candidates. What would be the expected behaviour of such a callback then? Should it fire after evaluating all the seeds, with results for all of them? Perhaps we could have a special flag/parameter in the callback parameters for the following events whenever the candidate is one of the seed candidates:
on_evaluation_start
on_evaluation_end
on_evaluation_skipped

> 2. on_valset_evaluated (or similar)

This is a great catch. @dbreunig, I think we should definitely have callback events for validation.

> 3. on_example_evaluated with iteration/candidate context

A simpler implementation can simply be to pass the inputs and outputs of "self.adapter.evaluate" as another parameter to the evaluation callbacks.

raveeshbhalla:
> > 1. on_seed_validated (or similar)
> 
> GEPA will soon allow users to provide multiple seed candidates. What would be the expected behaviour of such a callback then? Should it fire after evaluating all the seeds, with results for all of them? Perhaps we could have a special flag/parameter in the callback parameters for the following events whenever the candidate is one of the seed candidates: on_evaluation_start on_evaluation_end on_evaluation_skipped
>

Glad we're doing this! Generally, I want to be able to map all LM calls and their evals back to the actual program used, so whatever approach works. Always preferable for them to be mostly individually done, but depends on the rest of the architecture too.

LakshyAAAgrawal:
Thank you so much for the valuable PR, @dbreunig, and thanks @raveeshbhalla for your feedback as well.

Issue #162, opened 2026-01-13
KeyError / NoneType error in propose_new_texts when feedback is missing for a predictor


ZenWayne:
Description:
I encountered a bug in gepa/proposer/reflective_mutation/reflective_mutation.py when using the AllReflectionComponentSelector. 

When this selector is used, predictor_names_to_update includes all predictors within a module. However, if make_reflective_dataset does not generate feedback for every
single one of those predictors, the propose_new_texts function fails. It attempts to access a reflective dataset that doesn't exist in the mapping, leading to a crash.

Example Scenario:
 1. Predictors to update: predictor1, predictor2, predictor3 (e.g., all predictors in the module).
 2. Reflective datasets: Feedback is successfully generated only for predictor1 and predictor2.
 3. Execution: When the loop reaches predictor3, it fails at line 67 because the feedback dataset is missing.

Suggested Fix:
Add a check to skip predictors that do not have a corresponding reflective dataset.

PR #163, opened 2026-01-13
(fix)Fix crash when reflective feedback is missing for some predictors

ZenWayne:
Description:
  This PR addresses an issue in the reflective mutation proposer where the system would raise an error if a predictor targeted for update did not have corresponding feedback data.

Changes:
 - Modified gepa/proposer/reflective_mutation/reflective_mutation.py.
 - Added a safety check in propose_new_texts to verify if dataset_with_feedback is None (or missing).
 - If no feedback dataset is found for a specific predictor, the loop now correctly skips it and continues with the next predictor.

This ensures that components like AllReflectionComponentSelector can be used safely even if only a subset of predictors receives feedback during the optimization
process.

Verification:
Verified that adding if dataset_with_feedback is None: continue prevents the crash and allows the optimization to proceed for the remaining predictors.

Reference 
#162

### DIFF
--- a/src/gepa/proposer/reflective_mutation/reflective_mutation.py
+++ b/src/gepa/proposer/reflective_mutation/reflective_mutation.py
@@ -78,7 +78,9 @@ class ReflectiveMutationProposer(ProposeNewCandidate[DataId]):
                 continue
 
             base_instruction = candidate[name]
-            dataset_with_feedback = reflective_dataset[name]
+            dataset_with_feedback = reflective_dataset.get(name, None)
+            if dataset_with_feedback == None:
+                continue
             new_texts[name] = InstructionProposalSignature.run(
                 lm=self.reflection_lm,
                 input_dict={


LakshyAAAgrawal:
> This ensures that components like AllReflectionComponentSelector can be used safely even if only a subset of predictors receives feedback during the optimization process.

Will it not be better to use a different reflection component selector? The whole point of the reflection component selector is to identify which components to update, and it does get access to the same reflective_dataset.

LakshyAAAgrawal:
@ZenWayne, please let me know if this still doesn't address the problem, and I will be happy to merge this. Perhaps we may need to print a warning when the reflection dataset is not detected for a particular component.

ZenWayne:
actually i encounter this issue when using dspy to finetune composited prompt, i add some hook function to implement the finetuning for composited prompt, and use AllReflectionComponentSelector, i think i should add a custom selector support for dspy as you mention, i will close the issue cos it will not happened for dspy repo

PR #167, opened 2026-01-22
Fix a 1-line bug where cloudpickle wasn't used sometimes when it should

4gatepylon:
This causes state to fail to be saved after training with a serialization crash if you are using certain DSPY constructs that define types at runtime.


LakshyAAAgrawal:
Fixes #168

LakshyAAAgrawal:
Dear @4gatepylon,

This is a great bug catch. Thanks a lot!

PR #171, opened 2026-01-24
Fix MLflow logging, and add tracking for how many metric calls have been issued

chenmoneygithub:
This PR handles two things:

- Fix Mlflow logging, which was broken because the metrics dict contains string value, while MLflow enforces the metrics dict to only have numbers. 
- Track the number of total metric calls that have been issued in the experiment tracker. This is useful to track the total progress of GEPA optimization. Without tracking this information, users must read the live GEPA log to get the current progress.

### DIFF
--- a/src/gepa/logging/utils.py
+++ b/src/gepa/logging/utils.py
@@ -80,7 +80,7 @@ def log_detailed_metrics_after_discovering_new_program(
         "val_evaluated_count_new_program": coverage,
         "val_total_count": valset_size,
         "val_program_average": valset_score,
-        "metric_calls_issued": gepa_state.total_num_evals,
+        "total_metric_calls": gepa_state.total_num_evals,
     }
     if log_individual_valset_scores_and_programs:
         metrics.update(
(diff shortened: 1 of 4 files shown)


src/gepa/logging/utils.py
```diff
@@ -80,6 +80,7 @@ def log_detailed_metrics_after_discovering_new_program(
         "val_evaluated_count_new_program": coverage,
         "val_total_count": valset_size,
         "val_program_average": valset_score,
+        "metric_calls_issued": gepa_state.total_num_evals,
```
chenmoneygithub: "metric_calls_issued" tracks the total number of metric calls that have been issued. Please let me know if you prefer a different name.

LakshyAAAgrawal: "total_metric_calls" seems better to me but this is also okay

LakshyAAAgrawal:
Thanks a lot @chenmoneygithub, this is a great catch.

Could you please fix the typing issues, and I will go ahead and merge.

src/gepa/logging/utils.py
LakshyAAAgrawal: Could you also finalize this

LakshyAAAgrawal:
@copilot could you update all "metric_calls_issued" to "total_metric_calls".

src/gepa/logging/utils.py
chenmoneygithub: For sure, missed it, let me fix

LakshyAAAgrawal:
Thanks a lot @chenmoneygithub!

PR #231, opened 2026-02-25
(2/4) Persist OptimizationState via adapter_state in OptimizeAnythingAdapter

lukeleeai:
Summary
- Promote OptimizationState to the persistent state container for the optimize_anything adapter by adding best_evals_by_example field
- Implement get_adapter_state() / set_adapter_state() on OptimizeAnythingAdapter to sync with GEPAState at checkpoint boundaries
- Warm-start context (opt_state.best_example_evals) now survives resume — evaluators no longer lose accumulated best-eval history

Problem: _best_evals_by_example (the top-K best evaluations per example, used for warm-starting via OptimizationState) was adapter-local memory lost on resume. After reloading from a checkpoint, evaluators received empty
opt_state.best_example_evals.

Solution: Uses the adapter_state mechanism from #XXX (feature/add_adapter_state). The adapter stores its OptimizationState object in adapter_state["opt_state"], which gets pickled into gepa_state.bin at each checkpoint
and restored on resume.

Changes:
- OptimizationState — added best_evals_by_example: dict[str, list[dict]] field + __setstate__ for forward-compatibility when new fields are added
- OptimizeAnythingAdapter — replaced raw _best_evals_by_example dict with _opt_state: OptimizationState, added get/set_adapter_state()
- Evaluator API unchanged — opt_state.best_example_evals still works as before

### DIFF
--- a/src/gepa/adapters/optimize_anything_adapter/optimize_anything_adapter.py
+++ b/src/gepa/adapters/optimize_anything_adapter/optimize_anything_adapter.py
@@ -84,9 +84,11 @@ class OptimizeAnythingAdapter(GEPAAdapter):
         self.cache_mode = cache_mode
         self.cache_dir = Path(cache_dir) if cache_dir else None
 
-        # Track top-K best evaluations per example for warm-start support
-        self._best_evals_by_example: dict[str, list[dict]] = {}  # keyed by _example_hash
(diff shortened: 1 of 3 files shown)


LakshyAAAgrawal:
@lukeleeai does this PR need to be created with target main?

PR #248, opened 2026-03-12
feat: add edit mode for candidate proposal

### LINKED ISSUE #238: Edit mode, in place of fully rewriting the candidate

LakshyAAAgrawal:
## Summary
- Adds `InstructionEditSignature` that asks the LLM to make targeted edits instead of rewriting the entire instruction from scratch
- Preserves working parts of long prompts and uses fewer tokens
- Configurable via `proposal_mode="edit"` in `optimize()` or `ReflectionConfig(proposal_mode="edit")` in `optimize_anything()`
- Default remains `"rewrite"` for backward compatibility

**Usage:**
```python
# Via optimize()
result = optimize(..., proposal_mode="edit")

# Via optimize_anything()
result = optimize_anything(..., reflection=ReflectionConfig(proposal_mode="edit"))
```

Closes #238

## Test plan
- [ ] `uv run ruff check src/` passes
- [ ] `uv run pytest` passes
- [ ] Verify edit mode produces targeted changes vs full rewrites

### DIFF
--- a/src/gepa/api.py
+++ b/src/gepa/api.py
@@ -3,6 +3,7 @@
 
 import os
 import random
+import warnings
 from collections.abc import Sequence
 from typing import TYPE_CHECKING, Any, Literal, cast
 
@@ -341,6 +342,20 @@ def optimize(
     if cache_evaluation:
         evaluation_cache = EvaluationCache[RolloutOutput, DataId]()
 
+    # Warn if edit mode is combined with a custom template — the custom template
+    # won't contain SEARCH/REPLACE instructions, so edit mode silently falls back
+    # to rewrite mode.  Treat as rewrite to avoid confusion.
+    effective_proposal_mode = proposal_mode
+    if proposal_mode == "edit" and reflection_prompt_template is not None:
+        warnings.warn(
+            "proposal_mode='edit' with a custom reflection_prompt_template will use "
+            "rewrite mode. Edit mode requires SEARCH/REPLACE output format instructions "
+            "which are only present in the default template. Either remove the custom "
+            "template to use edit mode, or set proposal_mode='rewrite'.",
+            stacklevel=2,
+        )
+        effective_proposal_mode = "rewrite"
+
     reflective_proposer = ReflectiveMutationProposer(
         logger=logger,
         trainset=train_loader,
@@ -355,7 +370,7 @@ def optimize(
         reflection_prompt_template=reflection_prompt_template,
         custom_candidate_proposer=custom_candidate_proposer,
         callbacks=callbacks,
-        proposal_mode=proposal_mode,
+        proposal_mode=effective_proposal_mode,
     )
 
     def evaluator_fn(
--- a/src/gepa/strategies/instruction_proposal.py
+++ b/src/gepa/strategies/instruction_proposal.py
@@ -282,25 +282,25 @@ Rules:
 - To delete text, use an empty REPLACE section."""
 
     @classmethod
-    def run_with_metadata(
-        cls, lm: "LanguageModel", input_dict: Mapping[str, Any]
-    ) -> tuple[dict[str, str], str | list[dict[str, Any]], str]:
-        """Like ``run()``, but also returns the rendered prompt and raw LM output.
+    def _apply_edits_with_retry(
+        cls,
+        original_instruction: str,
+        lm_out: str,
+        lm: "LanguageModel",
+        max_retries: int = 1,
+    ) -> tuple[dict[str, str], str]:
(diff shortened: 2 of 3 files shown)


LakshyAAAgrawal:
@KE7 Could you please review

rishabh-16:
## Bug Report: Edit mode is silently inactive in two independent ways

I tested `proposal_mode="edit"` with `optimize_anything()` and found that edit mode **never actually activates**. There are two independent issues, either of which alone is sufficient to prevent edit mode from working.

### Issue 1: `_build_reflection_prompt_template()` overwrites the edit prompt

When `objective` or `background` is provided to `optimize_anything()` (which is the common case), line 1386 in `optimize_anything.py` replaces `config.reflection.reflection_prompt_template` with the output of `_build_reflection_prompt_template()`. This function **hardcodes** the output format to backtick blocks:

```python
# optimize_anything.py lines 587-594
sections.append("""
## Output Format

Provide ONLY the improved version within ``` blocks. The output must be a complete, 
drop-in replacement for the current component (whether it's a prompt, configuration, 
code, or any other parameter type).
Do not include explanations, commentary, or markdown outside the ``` blocks.""")
```

This template is then passed to `InstructionEditSignature` via `input_dict["prompt_template"]`. Since `prompt_template is not None`, `InstructionEditSignature.prompt_renderer()` uses it directly (line 98-100 of `instruction_proposal.py`) and **never falls through to `InstructionEditSignature.default_prompt_template`** — the one that contains the SEARCH/REPLACE instructions. The LLM never sees any SEARCH/REPLACE formatting guidance.

The LLM outputs a full rewrite in backtick blocks (as instructed). `_apply_edits()` finds zero `<<<<<<< SEARCH` markers, returns `None`, and the fallback at line 347-348 extracts from backticks — identical to rewrite mode.

**Evidence:** I dumped the `prompt` field from the proposals table of a running edit-mode job. The prompt contains "Provide ONLY the improved version within ``` blocks" and zero mentions of SEARCH/REPLACE. The `raw_lm_output` starts with `` ```text `` (full rewrite).

### Issue 2: `run_with_metadata()` bypasses `InstructionEditSignature.run()`

Even if Issue 1 were fixed, edit mode would still not work. The caller in `reflective_mutation.py:141` calls:

```python
result, prompt, raw_output = signature_cls.run_with_metadata(...)
```

`run_with_metadata()` is defined on the **base class** (`base.py:53-64`):

```python
@classmethod
def run_with_metadata(cls, lm, input_dict):
    full_prompt = cls.prompt_renderer(input_dict)
    lm_res = lm(full_prompt)
    lm_out = lm_res.strip()
    return cls.output_extractor(lm_out), full_prompt, lm_out
```

This calls `cls.output_extractor()` which does **backtick extraction** (inherited from `InstructionProposalSignature`). It does NOT call `cls.run()`.

But `InstructionEditSignature`'s SEARCH/REPLACE parsing, retry logic, and fuzzy matching all live in `InstructionEditSignature.run()` (lines 311-348). `InstructionEditSignature` does not override `run_with_metadata()`, so the entire edit pipeline is bypassed.

Even if the LLM correctly output SEARCH/REPLACE blocks, `run_with_metadata` would try to extract from backticks, likely returning garbage or the raw SEARCH/REPLACE text as the "new instruction."

### Summary

| Issue | What happens | Root cause |
|---|---|---|
| 1. Template override | LLM never sees SEARCH/REPLACE instructions | `_build_reflection_prompt_template()` hardcodes backtick output format, overwriting edit template |
| 2. Wrong method called | SEARCH/REPLACE parsing never runs | `run_with_metadata()` calls `output_extractor()` not `run()`, bypassing `InstructionEditSignature`'s edit logic |

Both issues must be fixed for edit mode to work. Fixing only one is insufficient.

### Suggested fixes

1. `_build_reflection_prompt_template()` should accept `proposal_mode` and emit SEARCH/REPLACE output format instructions when mode is `"edit"`.
2. `InstructionEditSignature` should override `run_with_metadata()` to call its own `run()` method (which has the SEARCH/REPLACE parsing + retry + fallback), and return the prompt and raw output alongside the result.

KE7:
Ahh this was what we previously spoke about. This should help with artifacts that are very long. Ty rishabh for the previous review. I see the edits were made to address them.

Basically, you introduce a new edit signature which asks an LLM to propose edit blocks of text as opposed to full rewrites. The edit blocks are then applied by `_fuzzy_find_and_replace` which tries 1) exact match, 2) whitespace-normalized match, and then 3) fuzzy line match. If there are edits that still failed, we do another LLM call. 

LGTM

PR #249, opened 2026-03-12
fix: filter non-numeric metrics from wandb/mlflow logging

### LINKED ISSUE #137: log_metrics for MLFlow raise excemption for string value
[Here](https://github.com/gepa-ai/gepa/blob/main/src/gepa/proposer/reflective_mutation/reflective_mutation.py#L134) we log dict new_instruction: str -> instruction: str. Then [here](https://github.com/gepa-ai/gepa/blob/main/src/gepa/logging/experiment_tracker.py#L107) exception is raised. Cause MLFlow accepts only float as a value to metric. Not a string

LakshyAAAgrawal:
## Summary
- Applies the same numeric-only filtering to `wandb.log()` that was already done for MLflow (line 108)
- Non-numeric values (strings, dicts, lists) were creating noisy charts in WandB dashboards

Closes #137

## Test plan
- [ ] `uv run ruff check src/` passes
- [ ] `uv run pytest` passes
- [ ] Verify WandB only receives numeric metrics

### DIFF
--- a/src/gepa/logging/experiment_tracker.py
+++ b/src/gepa/logging/experiment_tracker.py
@@ -96,7 +96,10 @@ class ExperimentTracker:
             try:
                 import wandb  # type: ignore
 
-                wandb.log(metrics, step=step)
+                # Filter out non-numeric values to avoid creating noisy string charts
+                numeric_metrics = {k: v for k, v in metrics.items() if isinstance(v, int | float)}
+                if numeric_metrics:
+                    wandb.log(numeric_metrics, step=step)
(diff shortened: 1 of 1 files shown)


LakshyAAAgrawal:
Closing — this change is now included in PR #243 which provides a more comprehensive fix: non-numeric data is logged via `log_table()` (structured wandb Tables / mlflow artifacts), and `log_metrics()` filters to numeric-only for both wandb and mlflow. Closes #137 via #243 instead.

PR #285, opened 2026-03-21
fix: wandb tables only show last row — accumulate across log_table calls

LakshyAAAgrawal:
## Problem

The `candidates` table (and all other tables: `valset_scores`, `valset_pareto_front`, `objective_scores`, `objective_pareto_front`) only shows a single row in wandb, even when multiple candidates were accepted during optimization.

### Root cause

Each `log_table()` call creates a fresh 1-row `wandb.Table` and logs it with `commit=False`:

```python
table = wandb.Table(columns=columns, data=data)  # 1 row
wandb.log({table_name: table}, commit=False)
```

`commit=False` queues the data in wandb's pending dict. When a second `log_table("candidates", ...)` call happens before the pending dict is flushed, the new 1-row table **overwrites** the previous one under the same key. Only the last row survives.

## Fix

Maintain a `_wandb_table_rows` dict that accumulates all rows per table name. Each `log_table()` call appends the new rows and logs the full growing table:

```python
# First call:  table has [row1]
# Second call: table has [row1, row2]
# Third call:  table has [row1, row2, row3]
```

The pending dict always contains the complete table, so no rows are lost regardless of commit timing.

## Verified

E2E test with `use_wandb=True`: https://wandb.ai/lakshyaaagrawal-university-of-california-berkeley/gepa-table-fix/runs/bnf5kon5

## Test plan
- [x] 370 tests pass
- [x] pyright: 0 errors, ruff: clean

### DIFF
--- a/docs/docs/guides/experiment-tracking.md
+++ b/docs/docs/guides/experiment-tracking.md
@@ -225,27 +225,114 @@ with open(local_path) as f:
 
 ---
 
-## Nesting Inside Existing Runs
+## Logging Into an Already-Active Run
 
-### WandB
+When GEPA is embedded inside a larger workflow that already has an active
+wandb/MLflow run, use `wandb_attach_existing=True` or `mlflow_attach_existing=True`
+to log GEPA metrics into it **without touching the run lifecycle**.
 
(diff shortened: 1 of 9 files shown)


LakshyAAAgrawal:
Closing — the table accumulation fix was already merged to main via PR #280 (which includes both key_prefix and table row accumulation). This PR has no remaining changes beyond a misplaced duplicate initialization.

Issue #288, opened 2026-03-23
feat: auto-inject score into side_info when not already present


LakshyAAAgrawal:
## Problem

The reflection LM only sees what's in `side_info` — the numeric `score` returned by the evaluator is used for acceptance/rejection and Pareto tracking but is **not automatically included** in the reflective dataset that the reflection LM reads.

This means if a user returns:

```python
def evaluator(candidate, example):
    result = run_model(candidate, example)
    return score, {"Input": example, "Output": result}
```

The reflection LM sees `Input` and `Output` but has **no idea what score the candidate got on this example**. It can't distinguish a 0.1 result from a 0.9 result. Users have to manually add `"Score": score` to their `side_info` dict to make this work, which is easy to forget and non-obvious.

## Current flow

In `OptimizeAnythingAdapter.make_reflective_dataset()` (line 526), the code iterates over `side_info.items()` and copies them into the reflective dataset. But `score` is a separate return value from the evaluator — it's never merged into `side_info`.

## Proposed fix

In `make_reflective_dataset()`, if the `side_info` dict for an example does not already contain a `"Score"` key (case-insensitive check to avoid overwriting user-provided scores), inject it automatically:

```python
for score, side_info in zip(scores, side_infos):
    record = {}
    # Auto-inject score if not already present
    has_score_key = any(k.lower() == "score" for k in side_info)
    if not has_score_key:
        record["Score"] = score
    for k, v in side_info.items():
        ...
```

This ensures the reflection LM always knows how well each example scored, even if the user didn't explicitly include it.

## Why this matters

The reflection LM's ability to diagnose failures depends on knowing which examples failed and by how much. Without the score, it treats all examples equally — a near-perfect 0.95 and a total failure 0.0 look the same. This is especially important for multi-example minibatches where some examples pass and others fail.

Issue #264, opened 2026-03-17
feat: built-in FeedbackBuilder utility for structured reflective datasets


LakshyAAAgrawal:
## Problem

Constructing rich, structured feedback for GEPA's reflective dataset requires custom boilerplate in every adapter. Dropbox [built a custom feedback construction pipeline](https://dropbox.tech/machine-learning/optimizing-dropbox-dash-relevance-judge-with-dspy) that composed: gap size and direction, human rationale, model reasoning, and anti-overfit constraints — all per example.

This pattern is common but users have to re-implement it from scratch.

## Proposed Solution

Ship a `FeedbackBuilder` utility that composes structured feedback records:

```python
from gepa.utils import FeedbackBuilder

builder = FeedbackBuilder(
    # Automatically adds "Predicted X but expected Y. Model rated N points higher/lower."
    include_score_diff=True,
    # Include the human rationale field from the example
    rationale_field="explanation",
    # Append global constraints to every record
    global_constraints=[
        "Do not copy specific keywords or examples from the training data.",
        "Do not change the fundamental task parameters (rating scale, output schema).",
    ],
)

def make_reflective_dataset(self, candidate, eval_batch, components_to_update):
    return builder.build(eval_batch, components_to_update)
```

## Context

- Dropbox blog: https://dropbox.tech/machine-learning/optimizing-dropbox-dash-relevance-judge-with-dspy
- Currently possible but requires boilerplate in every adapter's `make_reflective_dataset()`
- Addresses both the overfitting problem (#261) and task drift problem (#262) via `global_constraints`

Issue #262, opened 2026-03-17
feat: task definition preservation — prevent optimizer from changing task parameters


LakshyAAAgrawal:
## Problem

GEPA's reflection LM can drift the fundamental task definition — for example, changing a rating scale from 1-5 to 1-3, altering an output JSON schema, or modifying scoring criteria. This breaks the contract with downstream systems.

Dropbox [explicitly encountered this](https://dropbox.tech/machine-learning/optimizing-dropbox-dash-relevance-judge-with-dspy) when optimizing their Dash relevance judge and had to add constraints to prevent task parameter changes.

## Proposed Solution

Add an `invariant_check` mechanism to GEPA that validates proposed prompts before accepting them:

```python
result = gepa.optimize(
    ...
    prompt_invariants=[
        lambda old, new: "1-5" in new,          # rating scale preserved
        lambda old, new: '"score"' in new,       # output schema preserved
    ],
)
```

Options:
1. **`prompt_invariants`**: A list of `(old_prompt, new_prompt) -> bool` functions. If any returns `False`, the proposal is rejected (treated as no improvement).
2. **`frozen_substrings`**: A list of strings that must appear verbatim in every proposed prompt.
3. **`prompt_constraint`**: A string appended to every reflection call instructing the LM to preserve task parameters.

Option 3 is the simplest to implement and captures most real-world cases.

## Context

- Dropbox blog: https://dropbox.tech/machine-learning/optimizing-dropbox-dash-relevance-judge-with-dspy
- Workaround: users can currently inject this manually in `make_reflective_dataset()` — see FAQ
- Related to production safety: prompt changes that look like improvements on the training set may break downstream systems

Issue #261, opened 2026-03-17
feat: built-in anti-overfit guard to prevent prompt copying training data


LakshyAAAgrawal:
## Problem

GEPA's reflection LM can copy specific keywords, usernames, verbatim phrases, or domain-specific terms from training examples directly into the optimized prompt. This overfits the prompt to the training minibatch and hurts generalization.

Dropbox [explicitly encountered this](https://dropbox.tech/machine-learning/optimizing-dropbox-dash-relevance-judge-with-dspy) when optimizing their Dash relevance judge and had to manually append anti-overfit guardrails to every feedback record.

## Proposed Solution

Add a built-in `anti_overfit_guard` option to `gepa.optimize()` / `optimize_anything()` that:

1. **Detects verbatim copying**: After each reflection step, check whether the proposed prompt contains long substrings (e.g., n-grams of length ≥ 5) from the current training minibatch. If so, reject the proposal or log a warning.
2. **Injects guardrail instructions**: Automatically append an anti-overfit instruction to every reflective dataset entry when enabled — e.g., *"Do not copy specific examples, keywords, or phrases from the training data. Generalize to rules that apply broadly."*
3. **No-op by default**: Keep backward-compatible; guard is opt-in or at most a logged warning.

## Context

- Dropbox blog: https://dropbox.tech/machine-learning/optimizing-dropbox-dash-relevance-judge-with-dspy
- Workaround: users can currently inject this manually in `make_reflective_dataset()` — see FAQ
- This is the **highest priority** improvement because it's a silent failure mode (train score improves, production score doesn't)

PR #294, opened 2026-03-26
feat: add FeedbackBuilder utility for structured reflective datasets

rohitsandadi:
Implements `FeedbackBuilder` (#264). Addresses #261 (anti-overfit), #262 (task drift), #288 (score visibility in reflective dataset).

## Summary

`FeedbackBuilder` is a configurable utility in `gepa.utils` that replaces `make_reflective_dataset()` boilerplate. It replicates the existing side_info extraction logic, then layers on optional enrichments:

- **`include_score_diff`** — injects score + gap-from-perfect into each record's `Feedback` field, so the reflection LM always knows how examples scored (#288)
- **`rationale_field`** — extracts a named human-rationale field from side_info into a dedicated `Rationale` record field
- **`global_constraints`** — appends constraint text to every record, preventing the reflection LM from copying training data (#261) or drifting task definitions (#262)

When no `FeedbackBuilder` is configured, the adapter falls back to existing behavior.

## Changed files

- `src/gepa/utils/feedback_builder.py` — new: `FeedbackBuilder` dataclass with `build()` method
- `src/gepa/utils/__init__.py` — export `FeedbackBuilder`
- `src/gepa/adapters/optimize_anything_adapter/optimize_anything_adapter.py` — accept + delegate to `FeedbackBuilder`
- `src/gepa/optimize_anything.py` — add `feedback_builder` to `ReflectionConfig`, wire to adapter
- `tests/test_feedback_builder.py` — 13 unit tests

## Test plan

- [x] 13 tests covering base construction, each enrichment, combined features, import path, adapter integration
- [x] Full suite passes (383 tests)
- [x] `ruff check` + `pyright` clean

### DIFF
--- a/src/gepa/adapters/optimize_anything_adapter/optimize_anything_adapter.py
+++ b/src/gepa/adapters/optimize_anything_adapter/optimize_anything_adapter.py
@@ -29,6 +29,7 @@ from gepa.proposer.reflective_mutation.base import LanguageModel
 
 if TYPE_CHECKING:
     from gepa.optimize_anything import Candidate, OptimizationState, RefinerConfig, SideInfo
+    from gepa.utils.feedback_builder import FeedbackBuilder
 
 
 REFINER_PROMPT_TEMPLATE = """You are refining a candidate to improve its performance.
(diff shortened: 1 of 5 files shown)


ECNUACRush:
Hi! I was looking into #288/#264 and noticed this PR already covers the default Score injection path and part of the FeedbackBuilder direction.

Since this PR has been idle for a while, would maintainers prefer to keep this PR as the main path, or would a smaller follow-up PR focused only on #288's default Score injection be useful?

I’m happy to help with a narrowly scoped PR or tests/docs if that would make this easier to review, and also happy to defer to this PR if it is still the preferred direction.

Issue #277, opened 2026-03-19
feat: prompt mutation rate — control how much GEPA rewrites the prompt each iteration


LakshyAAAgrawal:
## Motivation

Users new to GEPA are often worried that the optimizer will completely rewrite their carefully crafted prompt in a single step — losing all the domain-specific wording, formatting, and constraints they already have. This is the **learning rate problem**: large updates are faster but riskier; small updates are safer but slower.

Currently GEPA's reflection LM can propose any change to the prompt — from a single-word tweak to a complete rewrite. There is no built-in way to constrain the magnitude of each mutation.

## Proposal

Add a `mutation_rate` parameter to `ReflectionConfig` (and the `gepa.optimize` API) that controls **how much of the prompt can change per iteration**, expressed as a fraction of the current prompt length:

```python
result = gepa.optimize(
    seed_candidate={"system_prompt": "Your carefully crafted prompt..."},
    ...
    mutation_rate=0.2,   # allow up to ~20% of the prompt to change per step
)
```

Or via `ReflectionConfig`:

```python
from gepa.optimize_anything import GEPAConfig, ReflectionConfig

config = GEPAConfig(
    reflection=ReflectionConfig(
        reflection_lm="openai/gpt-5",
        mutation_rate=0.2,   # 0.0 = no change, 1.0 = full rewrite (default)
    )
)
```

### How it would work

1. **Prompt-level instruction** (simplest, no parsing needed): when `mutation_rate < 1.0`, append an instruction to the reflection prompt template like:
   > *"Make targeted edits only. Your proposed prompt must retain at least 80% of the current prompt's content. Do not rewrite from scratch."*

2. **Post-hoc enforcement** (more robust): after the LM proposes a new prompt, compute the similarity to the old prompt (e.g. character-level diff ratio or token overlap). If the change exceeds the allowed rate, either:
   - Reject the proposal and ask for a smaller edit (retry with a stricter instruction)
   - Trim the diff to only apply the first N% of changes

### Suggested values

| `mutation_rate` | Behaviour |
|----------------|-----------|
| `1.0` (default) | Current behaviour — LM can rewrite entirely |
| `0.3` | Conservative — only ~30% of the prompt can change per step |
| `0.1` | Very conservative — small targeted edits only |
| `0.0` | No change allowed (effectively freezes the prompt) |

## Why This Matters

- **Trust**: users who have invested time in a prompt want confidence that GEPA will refine it, not replace it
[...]

PR #299, opened 2026-03-31
feat: add mutation_rate parameter to control mutation magnitude

rohitsandadi:
Implements `mutation_rate` (#277). Controls how much of the candidate the reflection LM can change per iteration — the "learning rate" analogy for text optimization.

## Summary

`mutation_rate` is a float (0.0–1.0) on `ReflectionConfig` and `gepa.optimize()` that appends a `## Mutation Constraint` section to the reflection prompt when < 1.0:

- **`1.0` (default)** — current behavior, LM can rewrite entirely
- **`0.3`** — conservative, constraint instructs LM to retain at least 70% of current content
- **`0.1`** — very conservative, small targeted edits only
- **`0.0`** — freezes the candidate, LM must return it as-is

When `mutation_rate` is not set, behavior is identical to before — no constraint is appended.

## Changed files

- `src/gepa/proposer/reflective_mutation/reflective_mutation.py` — accept `mutation_rate`, append constraint section to reflection prompt in `propose_new_texts()`
- `src/gepa/optimize_anything.py` — add `mutation_rate` field to `ReflectionConfig`, validate and wire to proposer
- `src/gepa/api.py` — add `mutation_rate` parameter to `optimize()`, validate and wire to proposer
- `tests/test_mutation_rate.py` — 6 unit tests for constraint text generation
- `tests/test_mutation_rate_integration.py` — 2 integration tests for ReflectionConfig wiring

## Test plan

- [x] 8 tests covering default behavior, retention percentages, freeze mode, generic language, prompt ordering
- [x] Full suite passes (74 tests, no regressions)
- [x] `ruff check` + `pyright` clean

### DIFF
--- a/src/gepa/api.py
+++ b/src/gepa/api.py
@@ -187,6 +187,10 @@ def optimize(
     if seed_candidate is None or not seed_candidate:
         raise ValueError("seed_candidate must contain at least one component text.")
 
+    # Validate mutation_rate
+    if not 0.0 <= mutation_rate <= 1.0:
+        raise ValueError(f"mutation_rate must be between 0.0 and 1.0, got {mutation_rate}")
+
     active_adapter: GEPAAdapter[DataInst, Trajectory, RolloutOutput] | None = None
     if adapter is None:
         assert task_lm is not None, (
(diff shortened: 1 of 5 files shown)


LakshyAAAgrawal:
When mutation rate is zero, we may even just return the string itself directly?

rohitsandadi:
Thanks for the suggestion, just updated the behavior in that case!

rohitsandadi:
@LakshyAAAgrawal updated my PR branch from feat/mutation-rate. Thanks!

PR #301, opened 2026-04-03
Unify evaluation caching: remove redundant adapter-level cache

lukeleeai:
## What changed

Before this PR, GEPA had **two independent evaluation caches** running simultaneously when `cache_evaluation=True`:

| | Core `EvaluationCache` (in `GEPAState`) | Adapter `_eval_cache` (in `OptimizeAnythingAdapter`) |
|---|---|---|
| Candidate key | Full 64-char SHA256 | Truncated to 16 chars |
| Example key | Stable `DataId` from DataLoader | Content hash (fragile for non-serializable objects) |
| Thread-safe | No | Yes |
| Persistence | Serialized with `gepa_state.bin` | Per-pair `.pkl` files in `fitness_cache/` |

The adapter cache always intercepted first (inside `_call_evaluator`), so the core cache **never saw a miss** — it was pure overhead. The two caches used different hash functions, making debugging and cross-referencing impossible.

**After this PR**, there is one cache: the core `EvaluationCache`. It now handles everything the adapter cache used to do (thread safety, disk persistence) plus new capabilities (trajectory caching).

### Adapter (`optimize_anything_adapter.py`)
- **Before**: 90+ lines of cache logic (`_eval_cache`, `_candidate_hash`, `_cache_key`, `_load_cache`, `_save_cache_entry`, disk `.pkl` logic, `threading.Lock`)
- **After**: `_call_evaluator()` is a 3-line pass-through. All caching handled by core.

### Core cache (`core/state.py` → `EvaluationCache`)
- **Before**: No thread safety, no disk persistence, no trajectory storage
- **After**:
  - `threading.Lock` on all reads/writes (with pickle-safe `__getstate__`/`__setstate__`)
  - `enable_disk_cache(path)` for write-through `.pkl` persistence to `{run_dir}/eval_cache/`
  - `trajectory` field on `CachedEvaluation` (with backward-compatible deserialization of old entries)
  - New `evaluate_batch_with_cache()` method with `require_trajectories` flag

### Proposer (`reflective_mutation.py`)
- **Before**: Always evaluated fresh with `capture_traces=True`, then wrote results to cache *without* trajectories. On next iteration, cache hit → but no trajectory → had to re-evaluate anyway.
- **After**: Calls `evaluate_batch_with_cache(require_trajectories=True)`. Cached entries *with* trajectories are reused directly. Entries without trajectories are treated as cache misses and re-evaluated. Results (including trajectories) are stored back so the next iteration benefits.

### `optimize_anything.py`
- **Before**: Passed `cache_mode`/`cache_dir` to adapter constructor
- **After**: Calls `evaluation_cache.enable_disk_cache()` on the core cache when disk mode is active. Adapter no longer needs cache config.

### Resume (`initialize_gepa_state`)
- **Before**: Disk cache was in adapter's `fitness_cache/` directory, loaded by adapter
- **After**: Disk cache is in `eval_cache/` directory, re-enabled on the deserialized `EvaluationCache` via `enable_disk_cache()`

## Public API

**No changes.** `cache_evaluation` and `cache_evaluation_storage` work exactly as before. The `eval_cache/` directory replaces `fitness_cache/` for disk persistence.

## Tests

- Added `@pytest.mark.llm_call` mock fixtures (per-file `_mock_litellm` autouse fixtures) so refiner/cache/best-evals tests run without real API keys — fixes the 6-hour CI timeout
- Added 12 new tests: disk write-through, trajectory caching, `evaluate_batch_with_cache` (all-miss/all-hit/partial/require_trajectories), thread safety (4 threads × 100 concurrent puts)


src/gepa/proposer/reflective_mutation/reflective_mutation.py
```diff
@@ -265,8 +265,15 @@ def execute_proposal(self, ctx: ProposalContext, state: GEPAState) -> ProposalOu
                 is_seed_candidate=ctx.is_seed_candidate,
             ),
         )
-        eval_curr = self.adapter.evaluate(ctx.minibatch, ctx.curr_prog, capture_traces=True)
-        total_evals += eval_curr.num_metric_calls if eval_curr.num_metric_calls is not None else len(ctx.subsample_ids)
+        if state.evaluation_cache is not None:
+            eval_curr, num_evals = state.evaluation_cache.evaluate_batch_with_cache(
```
Shangyint, line 269: Is this thread safe? execute proposals currently may run in multi-thread

src/gepa/optimize_anything.py
Shangyint, line 503: can we have some explanation here how is "auto" resolved?

Shangyint (commented):
LGTM! Left two minor comments

LakshyAAAgrawal:
Why is _save_to_disk is called outside the lock?
Similarly, why is _load_from_disk is called from enable_disk_cache without the lock?
evaluate_batch_with_cache accesses private _cache_dir from initialize_gepa_state. Could we ensure encapsulation?

src/gepa/optimize_anything.py
lukeleeai: done!

lukeleeai:
> Why is _save_to_disk is called outside the lock? Similarly, why is _load_from_disk is called from enable_disk_cache without the lock? 

Great questions! It's faster this way as I/O is relatively slower. And it's actually safe as everything we need is loaded/updated with lock, and we simply need to store/load them in a disk, and concurrent writes/reads for different entries go to / come from different files. 

> evaluate_batch_with_cache accesses private _cache_dir from initialize_gepa_state. Could we ensure encapsulation?

Good point! Updated.

src/gepa/optimize_anything.py
LakshyAAAgrawal: Is there any situation where run_dir is provided but the user would prefer to only keep in-memory?

src/gepa/core/state.py
```diff
@@ -129,6 +241,75 @@ def evaluate_with_cache_full(
 
         return outputs_by_id, scores_by_id, objective_by_id, len(uncached_ids)
 
+    def evaluate_batch_with_cache(
+        self,
+        candidate: dict[str, str],
```
LakshyAAAgrawal, line 208: We should support a list of candidate, as I was saying in Shangyin's PR earlier. So paired list of candidate and example ids.

PR #310, opened 2026-04-05
feat: add terrarium common eval infra for evolution/auto-research

Shangyint:
## Terrarium: Common eval infra for evolution/auto-research systems

Adds **terrarium**, a framework for defining benchmark tasks and running any evolution system against them with controlled evaluation budgets.

### Key design

All adapters — in-process (GEPA) or external (Claude Code) — go through a single `EvalServer` that enforces budget server-side. No evolution system can bypass or modify the budget counter.

```
In-process (GEPA)  ──► server.evaluate()  ──┐
                                             ├──► EvalServer (budget + eval_fn)
External (Claude Code) ──► POST /evaluate ──┘
```

### What's included

- **Core** — `Task`, `EvalServer`, `BudgetTracker`, `Adapter` protocol, CLI runner
- **4 starter tasks** — `circle_packing`, `optuna_blackbox`, `aime_math`, `arc_agi`
- **2 adapters** — GEPA (`optimize_anything`), Claude Code (black-box subprocess)

### Usage

```bash
uv pip install -e terrarium/
python -m terrarium circle_packing my_adapter.py --max-evals 100
```

Writing a new adapter is one method:

```python
class MyEvolver:
    def evolve(self, task, server, max_evals):
        score, info = server.evaluate(candidate)
        ...
        return Result(best_candidate=best, best_score=best_score)
```

See [`terrarium/README.md`](terrarium/README.md) for full docs on writing adapters and adding benchmarks.

### DIFF
--- a/terrarium/eval_server.py
+++ b/terrarium/eval_server.py
@@ -22,7 +22,7 @@ POST /evaluate_examples
     in parallel, respecting the concurrency limit. Each example = 1 budget tick.
     Response: {"average_score": 1.23, "scores": {"a": 0.9, ...}, "budget": {...}}
     For single-task problems (no dataset), behaves like /evaluate.
-    Status 429 when budget exhausted mid-evaluation (partial results returned).
+    Status 429 when budget is insufficient to evaluate all requested examples.
 
 GET /status
     Response: {"budget": {...}, "task": "<name>", "best_score": 1.23}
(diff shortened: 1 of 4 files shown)


terrarium/eval_server.py
```diff
@@ -0,0 +1,357 @@
...
+
+        This is the primary API for in-process adapters. Same budget counter
+        as the HTTP endpoint — there is only one counter.
+
+        Raises:
+            BudgetExhausted: When the budget has been used up.
+        """
+        self._eval_semaphore.acquire()
+        try:
+            self.budget.check()
```
lukedhlee, line 106: There can be a Time-Of-Check to Time-Of-Use issue when multiple threads are working in parallel. But it's not a serious issue, but just using a few more unnecessary evals

terrarium/runner.py
```diff
@@ -0,0 +1,181 @@
...
+    Returns:
+        Result from the evolution run.
+
+    Example::
+
+        result = terrarium.run("circle_packing", "my_adapter.py", max_evals=100)
+        print(result.best_score)
+    """
+    if isinstance(task, str):
+        task = get_task(task)
```
lukedhlee, line 52: This line uses `get_task` from a registry. But `tasks/aime_math` and `tasks/arc_agi.py` seems to only register a placeholder task without actual datasets? For example, [here](https://github.com/gepa-ai/gepa/pull/310/changes#diff-cb29a8b98d6604b25d5d3c85708ab07f6075c68467fe284f5a106a1b3e3f5ee2R121-R127).

lukedhlee:
Nice PR! Overall, this looks good, but adding some unit tests could be nice.

terrarium/eval_server.py
```diff
@@ -0,0 +1,357 @@
...
+        futures = {self._pool.submit(_eval_one, ex): ex for ex in examples}
+        budget_hit = False
+
+        for future in as_completed(futures):
+            eid, score, err = future.result()
+            if err == "budget_exhausted":
+                budget_hit = True
+                # Cancel remaining futures (best-effort)
+                for f in futures:
+                    f.cancel()
```
lukedhlee: `f.cancel()` only prevents queued futures. Already-running futures continue, call `budget.record()` and `_track()`, but their results are never collected.

terrarium/eval_server.py
```diff
@@ -0,0 +1,357 @@
...
+                break
```
lukedhlee: `break` exits the loop, so futures that already completed with valid scores before the budget-exhausted one are silently dropped. The `scores` dict is missing results, and `average_score` is computed over an incomplete subset.

PR #314, opened 2026-04-06
Parallel Proposals

Shangyint:
### Summary

Add parallel proposal execution to GEPA's reflective mutation proposer, enabling multiple propose-evaluate cycles to run concurrently within each optimization step. Also fixes MLflow logging to be thread-safe.

### Changes

**Parallel proposals** (`engine.py`, `reflective_mutation.py`, `optimize_anything.py`)
- Refactor `ReflectiveMutationProposer` into three phases: `prepare_proposal` (sequential) → `execute_proposal` (parallel-safe) → `apply_proposal_output` (sequential)
- Introduce `ProposalContext` (frozen snapshot of state for a single proposal) and `ProposalOutput` (deferred state updates returned by the proposal)
- Engine dispatches N proposals via `ThreadPoolExecutor` when `num_parallel_proposals > 1`, processes acceptances sequentially
- Add `num_parallel_proposals="auto"` mode: computes `max(1, max_workers // minibatch_size)` to fill the worker pool across concurrent proposals
- Default remains `1` (sequential) for backward compatibility
- Acceptance logic uses the configurable `AcceptanceCriterion` from main

**Thread-safe MLflow logging** (`experiment_tracker.py`)
- `mlflow.active_run()` is thread-local — parallel proposal threads were auto-creating new MLflow runs
- Fix: capture `run_id` at `start_run()` time, use `MlflowClient` with explicit `run_id` for metrics/params/artifacts
- Fall back to fluent API when `MlflowClient` is unavailable (test environments, no active run)

**Misc**
- Add `outputs/` to `.gitignore`
- Performance may drop, if this run under a fixed `max_metric_calls`

### DIFF
--- a/src/gepa/optimize_anything.py
+++ b/src/gepa/optimize_anything.py
@@ -1096,17 +1096,13 @@ class EvaluatorWrapper:
 
 def _resolve_num_parallel_proposals(
     value: int | Literal["auto"],
-    max_workers: int | None,
-    minibatch_size: int | None,
+    max_workers: int,
+    minibatch_size: int,
 ) -> int:
     """Resolve num_parallel_proposals, computing automatically if "auto"."""
     if isinstance(value, int):
         return value
-    workers = max_workers or (os.cpu_count() or 32)
-    mb = minibatch_size or 1
-    # Each proposal evaluates mb examples concurrently during its eval phases.
-    # N proposals × mb workers = total concurrent eval workers.
-    return max(1, workers // mb)
+    return max(1, max_workers // minibatch_size)
 
 
 def optimize_anything(
--- a/src/gepa/optimize_anything.py
+++ b/src/gepa/optimize_anything.py
@@ -483,6 +483,12 @@ class EngineConfig:
     parallel: bool = True
     max_workers: int | None = field(default_factory=lambda: os.cpu_count() or 32)
 
+    # Number of parallel proposal workers per optimization step.
+    # When > 1, multiple minibatches are sampled and proposed concurrently
+    # (each with its own evaluate-propose-evaluate pipeline), then acceptances
+    # are processed sequentially.
+    num_parallel_proposals: int = 1
+
     # Evaluation caching
     cache_evaluation: bool = False
     cache_evaluation_storage: CacheEvaluationStorage = "auto"
@@ -1585,6 +1591,7 @@ def optimize_anything(
         acceptance_criterion=acceptance_criterion_instance,
         use_cloudpickle=config.engine.use_cloudpickle,
         evaluation_cache=evaluation_cache,
+        num_parallel_proposals=config.engine.num_parallel_proposals,
     )
 
     # --- 15. Run optimization ---
--- a/src/gepa/optimize_anything.py
+++ b/src/gepa/optimize_anything.py
@@ -1608,8 +1608,8 @@ def optimize_anything(
         evaluation_cache=evaluation_cache,
         num_parallel_proposals=_resolve_num_parallel_proposals(
             config.engine.num_parallel_proposals,
-            config.engine.max_workers,
-            config.reflection.reflection_minibatch_size,
+            config.engine.max_workers or (os.cpu_count() or 32),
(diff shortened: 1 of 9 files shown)


src/gepa/optimize_anything.py
```diff
@@ -483,6 +483,16 @@ class EngineConfig:
...
 
+    # Number of parallel proposal workers per optimization step.
+    # When > 1, multiple minibatches are sampled and proposed concurrently
+    # (each with its own evaluate-propose-evaluate pipeline), then acceptances
+    # are processed sequentially.
+    # Set to "auto" to compute from max_workers and minibatch_size:
+    #   auto = max(1, max_workers // minibatch_size)
+    # Each proposal evaluates minibatch_size examples at a time, so this
+    # fills the worker pool across concurrent proposals.
+    num_parallel_proposals: int | Literal["auto"] = 1
```
LakshyAAAgrawal, line 494: @Shangyint this should ideally be a part of reflection config, since it only affects the reflectivemutationproposer?

LakshyAAAgrawal:
@Shangyint, there was also this PR: https://github.com/gepa-ai/gepa-optimize-anything-with-claude/pull/1/changes.

I recently created (https://github.com/gepa-ai/gepa/pull/279) in response to a user request. Building on this, I was actually planning that Adapter.evaluate() method signature should accept list[tuple[candidate, example]], instead of taking tuple(candidate, list[example]).

Once that is done, we could ensure that GEPA internally is NOT doing threading to make N parallel reflective proposals. Instead, we could do 3 steps:
1. Sample N ProposalContext
2. Call adapter.evaluate() with all candidates, datapoints that need to be evaluated
3. Call reflection LM with batch of N reflection prompts
4. Call adapter.evaluate() with validation for ALL accepted proposals.

And we won't even need to implement any threading/concurrency in GEPA's critical path.

Once this is done, it will enable us to then implement async as a supported primitive in GEPA, which has been a demanded feature by many: https://github.com/gepa-ai/gepa/pulls?q=is%3Apr+async

Shangyint:
nice! This all makes sense. I think step 2 are better than what we have currently because it definitely would support async well.

For step 3, I can see the point of not implementing `threading/concurrency in GEPA's critical path`, but this maybe needed to support external harness like claude code. IIUC you cannot send claude code a batch request.

> Call adapter.evaluate() with validation for ALL accepted proposals.

We probably need to do some evals on this. IMO this might need to be adaptive. For example, doing this with `iteration = 1` probably would use all the metric budget for experiments.

src/gepa/optimize_anything.py
Shangyint: I think this is better in EngineConfig: it's an engine orchestration parameter, not a proposer configuration. Proposer is not aware of this parameter (because we might still want proposer to be `some gepa state -> a new candidate`)

LakshyAAAgrawal: Makes sense!

LakshyAAAgrawal:
> For step 3, I can see the point of not implementing threading/concurrency in GEPA's critical path, but this maybe needed to support external harness like claude code. IIUC you cannot send claude code a batch request.

That is correct. My thinking is that GEPA internals should try to keep things sequential (and later async). adapter.evaluate() can choose how to parallelize, for example, DSPyAdapter uses DSPy's evaluator. OptimizeAnythingAdapter implements a threading based approach by default, but also lets the user provide their own batch_evaluator as per the PR above. So this way, we can fully avoid having any kind of threading related complications in GEPA's internals, to the extent possible.

For running parallel proposals, all language model providers support batch prompts (and from an inference perspective it will be even better). The only issue will be claude-code as proposer as you point out, in which case, I think claude-code-provider will basically have to implement the threading, not GEPA internals.

> We probably need to do some evals on this. IMO this might need to be adaptive. For example, doing this with iteration = 1 probably would use all the metric budget for experiments.

That is also correct, but all I am saying is, to the extent possible, if several things need to be evaluated, it is best to batch them together, and this ensures batching.

PR #325, opened 2026-04-11
feat(cost): track per-proposal reflection LM cost in GEPAState

lukeleeai:
## Summary

Track reflection LM cost (USD) for the default litellm reflection path and persist it in `gepa_state.bin`. Users can read the total from the run result or inspect per-proposal costs directly on state for plotting cost against metric-call budget.

## What gets tracked

Every proposal — accepted or rejected — contributes one `(metric_calls, reflection_cost)` point.

**On `GEPAState`** (persisted in `gepa_state.bin`, schema v6):

| field | shape | written when |
|---|---|---|
| `reflection_cost_by_candidate` | `list[float]`, parallel to `program_candidates` | a candidate is **accepted** |
| `reflection_cost_rejected` | `list[float]` | a proposal is **rejected** |
| `num_metric_calls_at_rejection` | `list[int]`, parallel to the rejected list | a proposal is **rejected** |

Accepted candidates already have their x-axis in the existing `num_metric_calls_by_discovery` list, so pairing `(num_metric_calls_by_discovery[i], reflection_cost_by_candidate[i])` gives the accepted points. The two new parallel rejected lists give the rejected points.

`GEPAState.total_reflection_cost` is a convenience property: `sum(by_candidate) + sum(rejected)`.

**On `GEPAResult`**:

```python
result = gepa.optimize(...)
print(f"${result.total_reflection_cost:.4f}")
```

A single scalar, populated from state at run end and round-tripped through `to_dict` / `from_dict`.

## How cost flows

`litellm.completion_cost(...)` is called inside `LM`, and the cost is returned through the reflection pipeline as a plain value — no shared counters, no hidden state.

**1. `LM.call_with_cost(prompt) -> (text, cost_usd)`** (`src/gepa/lm.py`)

The primary method. Runs `litellm.completion(...)`, asks `litellm.completion_cost(...)` for the USD cost, returns both. For self-hosted or unknown models where pricing data isn't available, cost is recorded as `0.0` — the call still succeeds.

`LM.__call__(prompt) -> str` is a thin wrapper that discards the cost and returns just the text, so existing `lm("prompt")` callers keep working unchanged.

**2. `run_with_metadata` returns a 4-tuple** (`src/gepa/proposer/reflective_mutation/base.py`)

`BasePromptSignature.run_with_metadata` returns `(result, prompt, raw_output, call_cost)`. It dispatches on whether the LM exposes `call_with_cost`:

```python
call_with_cost = getattr(lm, "call_with_cost", None)
if call_with_cost is not None:
    lm_res, call_cost = call_with_cost(full_prompt)
else:
    lm_res = lm(full_prompt)
    call_cost = 0.0
```

Plain-callable reflection LMs (lambdas, user classes) without `call_with_cost` fall through to the `lm(prompt)` path with `0.0` cost. The public `LanguageModel` protocol is unchanged.

**3. `propose_new_texts` sums cost across component calls** (`src/gepa/proposer/reflective_mutation/reflective_mutation.py`)

Returns `(new_texts, prompts, raw_lm_outputs, total_cost)`. `execute_proposal` unpacks it and attaches the total to the `CandidateProposal`:

```python
new_texts, prompts, raw_lm_outputs, reflection_cost = self.propose_new_texts(...)
proposal = CandidateProposal(..., reflection_cost=reflection_cost)
```

**4. Engine writes into state**

On accept, the reflective mutation path threads `proposal.reflection_cost` through `_run_full_eval_and_add` → `state.update_state_with_new_program`, which appends it to `reflection_cost_by_candidate`. On reject, the rejection branch appends `proposal.reflection_cost` to `reflection_cost_rejected` and the current budget `state.total_num_evals` to `num_metric_calls_at_rejection`.

## Custom reflection LMs

If the user passes a plain callable (not an `LM`), nothing breaks — `run_with_metadata` doesn't find `call_with_cost` via `getattr`, falls through to `lm(prompt)`, and every slot in `reflection_cost_by_candidate` / `reflection_cost_rejected` fills with `0.0`. Honest about what we don't know, no crash.

A user who *wants* cost tracking with their own callable can opt in by exposing a `call_with_cost(prompt) -> (text, cost_usd)` method on their class. That's the whole contract — no registration, no config.

## Reading the data

Post-run or mid-run (`state.save()` uses atomic `os.replace`, so a reader in another process never sees a torn write):

```python
from gepa.core.state import GEPAState

state = GEPAState.load("runs/my_run")

# single number
print(f"Total reflection cost: ${state.total_reflection_cost:.4f}")

# split by outcome
print(f"  accepted: ${sum(state.reflection_cost_by_candidate):.4f}")
print(f"  rejected: ${sum(state.reflection_cost_rejected):.4f}")

# timeline: cost-vs-budget step function
events = [
    *zip(state.num_metric_calls_by_discovery, state.reflection_cost_by_candidate, strict=True),
    *zip(state.num_metric_calls_at_rejection, state.reflection_cost_rejected, strict=True),
]
events.sort()
```

From a run result directly:
```python
result = gepa.optimize(...)
print(f"${result.total_reflection_cost:.4f}")
```


Shangyint (dismissed):
LGTM!

Shangyint:
Oh by the way @lukeleeai, can we add a stopper for reflection cost budget (similar to this https://github.com/gepa-ai/gepa/blob/d9e9ef832588b4e4aa34af157990524ed1bb2789/src/gepa/utils/stop_condition.py#L163-L164) and expose it in the optimize_anything api as well? should be an easy change.

lukeleeai:
> Oh by the way @lukeleeai, can we add a stopper for reflection cost budget (similar to this
> 
> https://github.com/gepa-ai/gepa/blob/d9e9ef832588b4e4aa34af157990524ed1bb2789/src/gepa/utils/stop_condition.py#L163-L164
> 
> ) and expose it in the optimize_anything api as well? should be an easy change.

Very good idea! Working on it

Shangyint:
LGTM, let's merge this!

LakshyAAAgrawal:
Question: Why are we adding a separate variable that we are tracking across all files (reflection_cost), instead of implementing an interceptor in the https://github.com/gepa-ai/gepa/blob/main/src/gepa/lm.py LM class, that we can use to wrap any reflection LM provided by the user. This can help count all things that litellm tracks, tokens in and out, cost in and out, total tokens, total cost, at VERY minimal code changes inside.

Shangyint:
I think this would allow us to track cost vs. iteration/candidate. If you just add it to gepa.lm, there's no way to do it?

Shangyint:
Plus, if we are going to support sessions, we need gepa state or session manager to track costs. Adding things to the LM class seems ad hoc.

LakshyAAAgrawal:
> If you just add it to gepa.lm, there's no way to do it?

Why not? You can always just read whatever is the current value of lm.cost or lm.tokens? The same can be done for session managers?

The current implementation means every new proposer, every new code path has to support more variable tracking. Having it managed in a class responsible for the work (like LM, or CodingAgentSessionManager) will ensure that once this code is written, any future code changes don't have to worry about it.

LakshyAAAgrawal:
I think we should refactor toward LM-level tracking. The core issue: the current approach threads cost through run_with_metadata → propose_new_texts → CandidateProposal → engine → GEPAState. Every new code path (ComBEE, sessions, coding agents, custom proposers) must manually wire this or cost goes untracked silently. Giving wrong cost tracking is worse than not tracking costs at all.

An interceptor on the LM class tracks cost automatically regardless of code path. Per-proposal granularity is still easy, just diff lm.cost before and after each proposal. For custom callables that aren't LM instances, we wrap them in a thin cost-tracking shim.

Flagging it so we don't accumulate more plumbing as new proposer types land.

We also had plans for multi-LLM proposers, which cannot be supported by this PR. I would really recommend fixing things properly as we implement them instead of leaving as future TODOs.

Shangyint:
I agree with you. Reverting. LM/SessionManager in the future is a better place to support it.

PR #326, opened 2026-04-11
Revert "feat(cost): track per-proposal reflection LM cost in GEPAState"

Shangyint:
Reverts gepa-ai/gepa#325


LakshyAAAgrawal:
I am just creating a PR fix btw.

PR #15, opened 2025-08-23
feat: improved the README.md for AnyMaths; fixed AnyMathsAdapter code; provided example training and dataset preparation; added eval results and sample eval code

egmaminta:
#### Summary
This PR improves the `AnyMaths` example end-to-end by fixing the adapter code (I found a bug that only outputs 1 Pareto val set), adding dataset preparation utilities, providing a runnable sample training script, and updating the `AnyMaths` README with clear usage guidance.

#### Changes
- Docs
    - Updated `README.md` with setup, configuration, and run instructions.
- Adapter
    - Fixed `AnyMaths` adapter to correctly handle multi-item evaluation (no more assertion error when valset > 1).
    - Ensured batched/iterative evaluation returns consistent metrics and aggregates correctly.
- Dataset preparation
    - Added dataset loading and splitting logic supporting:
        - `"openai/gsm8k"` (train/test) with deterministic shuffling and derived solutions.
        - `"MathArena/aime_2025"` with aligned train/test preprocessing.
    - Provides train/val split from train; expands a small test set for demo runs.
- Sample training
    - Added `train_anymaths.py`:
        - CLI args for model, API base, budget, reflection LM, minibatch size, and seed.
        - Uses `gepa.optimize` with `AnyMathsAdapter`.
        - LiteLLM-based reflection LM hook.
        - Small default subset sizes for quick demos.
- Benchmark results
    - Added `openai/gsm8k` - `test` split results of `ollama/qwen3:4b`. (Before GEPA: 18% accuracy; After GEPA: **23%** accuracy; Budget used: 5).
    - Added a sample eval script for `openai/gsm8k` (see `src/gepa/examples/anymaths-bench/eval_default.py`).

#### Why
- The adapter previously triggered an assertion when `valset > 1`. This is non-ideal for real validation and made experiments brittle.
- Providing dataset prep and a runnable script lowers the barrier to trying AnyMaths with GEPA.

#### How to use
- Start an Ollama server (or compatible endpoint) and ensure the requested models are available.
- Run the training script (example):
    - Module: `python -m gepa.examples.anymaths-bench.train_anymaths --model_name ollama/qwen3:4b --api_base http://localhost:11434 --anymaths_dset_name openai/gsm8k --budget 500`
- Adjust `--reflection_lm`, `--reflection_minibatch_size`, and `--seed` as needed.

#### Notes
- Defaults limit train/val/test to 10 items for fast demos; expand as desired.
- No breaking API changes expected outside the `AnyMaths` example path.

[...]

PR #346, opened 2026-04-29
feat: add engine-pluggable optimize_anything

Shangyint:
Porting terrarium's multi-backend into gepa. WIP

Introduces a unified Python API that dispatches a single optimization call to any compatible backend (GEPA, Claude Code, Meta-Harness, …) over a shared eval-server contract:

    from gepa.omni import optimize_anything, Task
    result = optimize_anything(
        task=Task(name="t", initial_candidate="...", objective="..."),
        evaluate=my_eval_fn,
        backend="gepa",
        config={"reflection": {"reflection_lm": "openai/gpt-5"}},
        max_evals=200,
        max_token_cost=5.0,
    )

Design choices that drove the layout:
- Task is pure data (name, initial_candidate, objective, background, train/val/test). No bundled eval_fn — code stays in Python.
- evaluate is a required, separate parameter; the eval server wraps it and is the single budget/concurrency choke point.
- One Backend protocol — no Adapter sublayer. Backends call server.evaluate (in-process) or POST to server.url (subprocess).
- max_token_cost is a single top-level field; the api threads it into whichever per-backend knob enforces it (gepa: max_reflection_cost; claude_code/meta_harness: --max-budget-usd).
- No Hydra inside gepa.omni — users call the function directly.

Results:
| Task | Task type | Budget | Baseline | [GEPA](src/gepa/oa/engines/gepa.py) | [autoresearch](src/gepa/oa/engines/autoresearch.py) | [Meta-Harness](src/gepa/oa/engines/meta_harness.py) |
| --- | --- | --- | --- | --- | --- | --- |
| Frontier-CS | Program | $20 | 7.72 | 43.8 | **55.4** | 50.9 |
| Can't-Be-Late | System | $20 | -96.48 | -91.12 | -89.52 | **-88.97** |
| Cloudcast | System | $20 | 0.00520 | 0.00900 | 0.00917 | **0.00919** |
| FiNER | Prompt | 4000 evals | 40.0% | **49.3%** | 47.3% | 40.7% |
| Formula | Prompt | 4000 evals | 77.3% | 78.7% | **87.3%** | 76.0% |
| LiveBench-Math | Prompt | 4000 evals | 80.9% | 83.0% | 81.4% | **86.3%** |


uv.lock
LakshyAAAgrawal, line 1: Revert this file change.

src/gepa/api.py
```diff
@@ -85,6 +85,7 @@ def optimize(
     track_best_outputs: bool = True,
     display_progress_bar: bool = False,
     use_cloudpickle: bool = False,
+    write_agent_state: bool = False,
```
LakshyAAAgrawal, line 91: If given a run_dir, we should just default to writing agent state? I am not sure this needs to be a flag. Thoughts?

LakshyAAAgrawal: Perhaps the agent state is too big? Can you provide some typical numbers on how large the agent state is across big GEPA runs?

src/gepa/optimize_anything.py
```diff
@@ -462,6 +462,14 @@ class EngineConfig:
...
     use_cloudpickle: bool = True
     track_best_outputs: bool = True
+    # Write an agent-readable directory tree (``iterations/NNNNN/`` +
+    # ``pareto/``) under ``run_dir`` on every save. Each loop iteration —
+    # accepted or rejected — gets its own directory with ``meta.json``,
+    # ``components/``, ``trace.json`` (with before/after scores and
+    # trajectories), plus ``val_scores.json`` + ``outputs/`` + ``trajectories/``
+    # for accepted ones. The seed is pinned at ``iterations/00000/``. Default
+    # off.
+    write_agent_state: bool = False
```
LakshyAAAgrawal: Same comment as on api.py. Let's discuss in that other thread. Keeping this just for flag.

src/gepa/core/state.py
```diff
@@ -33,6 +34,14 @@ def _candidate_hash(candidate: dict[str, str]) -> CandidateHash:
     return hashlib.sha256(json.dumps(sorted(candidate.items())).encode()).hexdigest()
 
 
+_COMPONENT_NAME_SANITIZER = re.compile(r"[^A-Za-z0-9_.-]")
```
LakshyAAAgrawal, line 38: Should . be excluded as well?

src/gepa/core/state.py
```diff
@@ -345,6 +385,257 @@ def save(self, run_dir: str | None, *, use_cloudpickle: bool = False) -> None:
...
+            self._save_agent_directory(run_dir)
+
+    def _save_agent_directory(self, run_dir: str) -> None:
+        """Write agent-readable state as a directory of small files.
+
+        Partitions the run's state into per-iteration subdirs so an agent
+        (e.g. Claude Code) can navigate a single timeline — both accepted
+        candidates *and* rejected proposals live under ``iterations/NNNNN/``.
+        ``iterations/00000/`` is always the seed (``candidate_idx=0``); subsequent
+        directories correspond to ``state.i + 1`` (the loop iteration that
```
LakshyAAAgrawal: What happens when we introduce concurrency/parallelism to GEPA, and state.i+1 is not unique?

src/gepa/oa/proposers/claude_code_agent.py
```diff
@@ -0,0 +1,815 @@
...
+            "--print",
+            wrapper,
+            "--output-format",
+            "json",
+            "--model",
+            self.model,
+            "--session-id",
+            session_id,
+            "--permission-mode",
+            "bypassPermissions",
```
Benzhang2004: **Conflicting `--permission-mode` flags.** This hardcodes `--permission-mode bypassPermissions`, but on line 702 `claude_settings_args(jail_root)` is appended, which adds a *second* `--permission-mode default` (`sandbox.py:224-225`). That `default` mode is exactly what the sandbox docstring says is **required** for the `permissions.allow` file-tool whitelist to enforce (`sandbox.py:174-175`, `212-216`).

So the two flags directly contradict each other and the documented tool-layer confinement becomes ambiguous (which one wins depends on CLI arg precedence). The OS-level Seatbelt/bwrap jail still applies, but the tool-layer whitelist may be silently disabled. Please reconcile to a single mode and make the intent explicit.

src/gepa/omni/registry.py
```diff
@@ -0,0 +1,74 @@
+"""Registries for backends and tasks.
+
+Backends register a string name → class; the api resolves ``backend="gepa"``
+to the class and calls ``cls.from_config(omni_config)``.
```
Benzhang2004: **Stale doc contract.** This says backends are built via `cls.from_config(omni_config)`, but the actual construction path calls `cls(config)` (`api.py:204`), matching the `Backend.__init__(self, config)` protocol (`backend.py:43`). No `from_config` classmethod exists. A backend author following this docstring would write the wrong constructor. Please update it to `cls(omni_config)`.

src/gepa/core/state.py
Shangyint: This would be a problem, but also the fix depends on how we decided to introduce concurrency/parallelism . So I'd defer the change.

src/gepa/oa/proposers/claude_code_agent.py
Shangyint: thanks resolved.

src/gepa/oa/engines/autoresearch.py
```diff
@@ -0,0 +1,710 @@
+"""AutoResearch engine: black-box optimization via a Claude Code subprocess.
```
LakshyAAAgrawal, line 1: How confident are we that this autoresearch implementation matches/is close to Karpathy's? Does he use ralph terminology?

LakshyAAAgrawal:
But it pays for that breadth with a hard breaking change to gepa.optimize_anything that's far larger than the PR description suggests. The new API takes over the existing module's namespace; EngineConfig, GEPAConfig, ReflectionConfig, MergeConfig, RefinerConfig, TrackingConfig etc. all disappear; ~25 in-repo docs/blog/examples/tutorials break on merge; the deprecation warning never fires for the most common case (ImportError on the moved symbols).

The fix is essentially free: the new module gepa.oa already exists and is half-set-up to be the canonical home for the new API. Put optimize_anything at gepa.oa.optimize_anything, leave gepa.optimize_anything as a re-export shim of gepa.legacy_optimize_anything, and you get 100% of the new functionality with 0% of the breakage. The PR is already using exactly this strategy in reverse (move old → legacy_*) — just flip which name is the "canonical" one.

src/gepa/core/engine.py
LakshyAAAgrawal, line 1: https://github.com/gepa-ai/gepa/blob/631403e7d0cdd6a42823c356ea932d5a043d316d/src/gepa/core/engine.py#L195-L195

Are these consistent?

src/gepa/oa/engines/autoresearch.py
Shangyint: very close. No ralph loop in the original impl, but we decided this is an easy and painless improvement

src/gepa/core/engine.py
Shangyint: yes. https://github.com/gepa-ai/gepa/blob/631403e7d0cdd6a42823c356ea932d5a043d316d/src/gepa/core/engine.py#L278

src/gepa/core/state.py
LakshyAAAgrawal: I am wondering if there's anything else to anchor on? Simply a GUID/UUID based schema? Like even when we use non-GEPA engine backends which already support parallelism (like claude-code with agent swarms), what do we expect will happen there?

LakshyAAAgrawal: Basically I am trying to think of doiung the minimal amount of changes to ensure we can support agent swarms and such things natively (because they will not be implemented too far away in the future, and we will struggle with that question within weeks of timeline).

LakshyAAAgrawal: A UUID/GUID based technique could potentially be adopted pretty easily right now?

src/gepa/core/state.py
```diff
@@ -345,6 +385,257 @@ def save(self, run_dir: str | None, *, use_cloudpickle: bool = False) -> None:
...
+            f"{base}/val_scores.json",
+            {str(k): v for k, v in self.prog_candidate_val_subscores[0].items()},
+        )
+        self._save_components_dir(run_dir, base, self.program_candidates[0])
+
+    def _save_proposal_iteration_dirs(self, run_dir: str, cand_to_iter: dict[int, int]) -> None:
+        """Write one ``iterations/NNNNN/`` per loop trace entry.
+
+        Covers both accepted and rejected proposals. Iteration id is
+        ``state.i + 1`` so the seed keeps id ``0``.
```
LakshyAAAgrawal: Could these be one method?

LakshyAAAgrawal: This class is already very complex. Should use DRY to the extent possible.

src/gepa/core/engine.py
```diff
@@ -159,10 +223,35 @@ def _evaluate_on_valset(
         valset = self.valset
         assert valset is not None
 
-        val_ids = self.val_evaluation_policy.get_eval_batch(valset, state)
+        val_ids = list(self.val_evaluation_policy.get_eval_batch(valset, state))
+
+        # When write_agent_state is on, bypass the cache so we can capture
+        # trajectories (which the cache does not store).
```
LakshyAAAgrawal, line 269: Isn't a more principled fix to store these things in the cache rather than bypass? Also I think there's a separate optimize_anything cache and GEPA cache. Is GEPA cache even useful? if not, we should disable and remove it entirely. If it is useful, let's use it here by adding relevant data to it.

src/gepa/gskill/gskill/train_optimize_anything.py
```diff
@@ -25,7 +25,7 @@
 from gskill.experiment_logger import ExperimentLogger, set_logger
 from gskill.swe_fitness_fn import create_swe_fitness_fn
 
-from gepa.optimize_anything import EngineConfig, GEPAConfig, ReflectionConfig, TrackingConfig, optimize_anything
+from gepa.legacy_optimize_anything import EngineConfig, GEPAConfig, ReflectionConfig, TrackingConfig, optimize_anything
```
LakshyAAAgrawal: If we are confident on the parity/perf of the new version, perhaps better to migrate to new one.

Shangyint:
@LakshyAAAgrawal iteration idx fixed. now everything will be uuid'd. thx!

LakshyAAAgrawal:
I had pointed my claude-code to refer to this PR, and set up meta-harness on some task I was doing locally. However, that implementation didn't add support for side info. Worth checking if all of that wiring works for all engines or not.

src/gepa/optimize_anything.py
```diff
@@ -1,1644 +1,289 @@
...
-
-    seed_candidate  →  evaluate  →  reflect on ASI  →  propose  →  repeat
-                          ↑                                  |
-                          └──────────────────────────────────┘
-
-Three optimization modes
-------------------------
-``optimize_anything`` unifies three optimization paradigms under one
-declarative API, determined by whether you provide ``dataset`` and ``valset``:
+"""Public ``optimize_anything`` API.
```
LakshyAAAgrawal, line 1: The main optimize_anything API should be a properly typed API, with detailed documentation explaining every arg.

If the issue is about backwards compatibility, the solution could be

def optany(
    new_arg1: Type1,
    new_arg2: Type2,
    *args,
    **kwargs
):
    if len(args) > 0 or len(kwargs)> 0:
       # print warning
       # Typecast args and kwargs as per old optany types
       # call optany

src/gepa/optimize_anything.py
```diff
@@ -1,1644 +1,289 @@
...
-    - :func:`optimize_anything` — main entry point
-    - :data:`SideInfo` — type alias for evaluation diagnostics (ASI)
-    - :class:`Evaluator` — evaluator protocol
-    - :class:`OptimizationState` — context injected into evaluators
-    - :class:`GEPAConfig`, :class:`EngineConfig`, :class:`ReflectionConfig`,
-      :class:`RefinerConfig`, :class:`MergeConfig`, :class:`TrackingConfig`
-    - :func:`log`, :func:`get_log_context`, :func:`set_log_context` — in-evaluator logging
-    - :func:`make_litellm_lm` — convert a model name string to a callable
-    - :class:`Image` — wrapper for including images in side_info (VLM reflection)
+Use :func:`optimize_anything_with_server` when an outer system already owns an
```
LakshyAAAgrawal: Didn't understand the meaning of this sentence. The key public facing API file should have very good documentation.

src/gepa/optimize_anything.py
```diff
@@ -1,1644 +1,289 @@
...
-   Evaluator is called without ``example``.
-   *E.g. circle packing, blackbox mathematical optimization.*
+The API has three pieces:
 
-2. **Multi-Task Search** (``dataset=<list>, valset=None``):
-   Solve a batch of related problems with cross-task transfer.  Insights
-   from solving one help solve the others.  Evaluator is called per-example.
-   *E.g. CUDA kernel generation for multiple PyTorch operations,
-   multi-aspect SVG optimization.*
+- :class:`Task` describes what to optimize.
```
LakshyAAAgrawal: I didn't understand what purpose does introducing Task as a class, introducing indirection serve. All of the arguments that Task takes were earlier first class arguments to optimize_anything, so semantically nothing changed, except that the user now has to wrap it in a Task(...) call. This isn't better even from the perspective of a coding agent which will have to read 2 files?

src/gepa/oa/eval_server.py
```diff
@@ -0,0 +1,575 @@
...
+    """
+    if isinstance(item, Mapping):
+        if "id" in item:
+            return str(item["id"])
+    elif hasattr(item, "id"):
+        return str(item.id)
+    return fallback
+
+
+class EvalServer:
```
LakshyAAAgrawal, line 62: Should add support for https://github.com/gepa-ai/gepa/pull/279 ? Batch evals that can run across candidates and examples.

LakshyAAAgrawal: Is there any built-in safety that prevents access from outside machine, or over the network, by default?

LakshyAAAgrawal: This file refers to "test" multiple times. So far GEPA hasn't had the concept of a test, do the other engines have it?

src/gepa/optimize_anything.py
Shangyint: removed

src/gepa/optimize_anything.py
Shangyint: I think no backward compatibility is probably okay. if new_arg1 is required we are essentially break this backward compat!

src/gepa/optimize_anything.py
Shangyint: Good call.  fixed.

Shangyint:
@LakshyAAAgrawal also removing the `metadata` arg in optimize_anything, it is mergable into engine config (only engine consumes them)

src/gepa/optimize_anything.py
```diff
@@ -1,1644 +1,277 @@
...
-
-def _build_reflection_prompt_template(objective: str | None = None, background: str | None = None) -> str:
-    """
-    Build a reflection prompt template dynamically based on provided objective and background.
-
-    Only includes sections that have content, ensuring the prompt feels natural
-    regardless of which optional parameters are provided.
+def optimize_anything(
+    *,
+    name: str,
```
LakshyAAAgrawal: Why is name a mandatory arg? We can make a UUID/random name on the fly if the user doesn't provide one? Or a path-datetimestamped name? Similar to how claude-code creates sessions.

LakshyAAAgrawal: Can we put initial_candidate back to seed_candidate? Why is initial_candidate now a mandatory arg even though we support seedless mode in optimize_anything?

LakshyAAAgrawal: I had asked in a separate thread: Is the support for test_set standard to provide? It hasn't been there in GEPA, we only had train and val set. In fact, some of the faculty didn't even like that there was separate train/val, saying the system can produce those internally. Just thinking out loud what should be the ideal here.

LakshyAAAgrawal: "background: Long-form context — problem statement, evaluation rules,
            domain notes. Surfaced verbatim by every engine."

background should ideally not be surfaced every iteration. GEPA is naive in that sense. But maybe for other engines especially the ones using coding agents, we can write background into a file, and point the agent to the file?

LakshyAAAgrawal: "name" can potentially be subsumed in config too. Or at least become the last argument, just before config.

Shangyint: I agree to all points here except "Is the support for test_set standard to provide?". 

My experience is that most of the time, I have to add support for test_set in the call site of optimize_anything. Thus I think it'd be better to have it for convinence. About val/train: I think it's okay to merge it and gepa should make the split. Wdyt?

Shangyint: As for background, I think we might even get rid of the concept as a whole? Do we know for sure that separating it from the objective would be beneficial?

Shangyint: But again, merging val/train would make it hard for controlled experiments (even less info is surfaced). I think the size of the optany arg number is reasonable, so we do not need to trim.

LakshyAAAgrawal:
Did a full second-pass review against the current head (`e700b6e`). I worked through every prior inline comment + my issue-level comment, then reviewed the new code against the patterns from your past merged-PR reviews (especially #100 / dynamic validation — same lens: don't break end users, prefer protocols, push back on new indirection, question removed assumptions, DRY, concurrency-ready). Findings below, with criticality-ordered list at the bottom.

## Status of your prior comments

| # | Comment | Status | Notes |
|---|---|---|---|
| 1 | **`uv.lock` — revert** | ❌ not addressed | Diff is now 1081 inserts / 567 deletions vs main. Same comment as on every other PR you've reviewed; this one keeps slipping. |
| 2 | **`api.py:88` write_agent_state — should default with `run_dir`** | ❌ not addressed | Still a `bool = False` flag in both `api.py:88` and `optimize_anything.py:472`. No numbers provided on agent-state size either. |
| 3 | **`state.py:38` — should `.` be excluded too?** | ❌ not addressed | Regex still `[^A-Za-z0-9_.-]`. Leading `.` would produce a hidden file on Unix; arguably you do want it banned. |
| 4 | **`state.py:399` — state.i+1 uniqueness under concurrency / agent swarms / UUID** | ✅ addressed | `new_iteration_id() = uuid.uuid4().hex[:8]`, `SEED_ITERATION_ID = "seed"`, `iteration_ids_by_candidate_idx` lives on state. Each entry now anchors on a random short hex id, not `state.i+1`. |
| 5 | **`autoresearch.py:1` — match Karpathy / ralph terminology** | ⚠️ partial | Code now uses "ralph" naming throughout (`RALPH_CONTINUE_PROMPT`, `_RALPH_MIN_HEADROOM_USD`, `ralph: bool` config knob) and the docstring says "in the spirit of karpathy/autoresearch". Your specific "how confident" ask doesn't have an answer in the PR though — no reference link, no parity test. |
| 6 | **`engine.py:1` — "are these consistent"** (linking L195) | ⚠️ unclear | L195 today is `_write_agent_iteration_files`. Hard to tell what the consistency check was vs without the original line you pointed at; need the author to explicitly reply. |
| 7 | **`state.py:466` — DRY, could these be one method** | ⚠️ partial | The save methods have an internal `write` helper now (`_save_iteration_dirs` lines 469-489). Some consolidation; not the full collapse the comment was pointing at. |
| 8 | **`engine.py:238` — store in cache rather than bypass** | ❌ not addressed | Code at L230-250 still has the comment "*When write_agent_state is on, bypass the cache so we can capture trajectories (which the cache does not store)*". Your suggestion was to extend the cache to store trajectories, not bypass. |
| 9 | **`gskill/train_optimize_anything.py:28` — migrate to new API** | ❌ not addressed | Still imports from `gepa.legacy_optimize_anything`. Probably acceptable if intentional, but the original question "are we confident on parity?" still doesn't have a yes. |
| 10 | **`optimize_anything.py:1` — typed API + backwards-compat wrapper** | ✅ addressed | Signature now mirrors legacy: `optimize_anything(seed_candidate, *, evaluator, dataset, valset, objective, background, test_set, config)`. Full docstring per arg. The wrapper-with-warning pattern you sketched isn't there (and no warning fires if user passes `config=GEPAConfig(...)` — see C2 below), but the named-arg compat is good. |
| 11 | **`optimize_anything.py:27` — confusing sentence** | ✅ addressed | Removed in `0750158`. |
| 12 | **`optimize_anything.py:5` — Task class indirection** | ✅ addressed | Task dissolved into top-level args (`9bb69f2`). Task is still constructed internally but no longer in the user's mental model. |
| 13 | **`eval_server.py:62` — support PR #279 batch evals** | ❌ not addressed | `evaluate_examples` exists for "one candidate, many examples" but not the candidate×example matrix from #279. |
| 14 | **`eval_server.py:62` — outside-machine safety** | ✅ addressed | `ThreadingHTTPServer(("localhost", port), Handler)` at L358. Bound to loopback only. (Note: `ThreadingHTTPServer` also fixes my earlier concurrency concern — good.) |
| 15 | **`eval_server.py:62` — "test" concept** | ⚠️ partial | `test_set` is still present. Now documented as opt-in (skipped if absent), and **sealed at the HTTP layer** so HTTP callers/agents can't see it. The conceptual question — *should GEPA have test at all?* — still open. |
| 16 | **`optimize_anything.py:67` — name mandatory?** | ✅ addressed | `config.name: str \| None = None`; `_default_run_name(engine)` generates `{engine}-{uuid8}-{ts}` when omitted. |
| 17 | **`optimize_anything.py:67` — initial_candidate → seed_candidate** | ✅ addressed | Renamed everywhere; also `seed_candidate: str \| None = None` supports seedless mode. |
| 18 | **`optimize_anything.py:67` — background surfaced every iteration is naive** | ❌ not addressed | `Task.background` docstring still says *"Surfaced verbatim by every engine"*. Your specific suggestion (write background to a file, point coding-agent engines at the file) isn't implemented for any engine. |
| 19 | **`optimize_anything.py:67` — name in config or last arg** | ✅ addressed | Now lives on `OptimizeAnythingConfig.name`, not as a function arg. |
| 20 | **Issue-level 2026-06-18 — backwards compat / namespace** | ⚠️ partial | The function signature now matches legacy positional/keyword shape, so most legacy call sites won't break on the function call itself. BUT `from gepa.optimize_anything import GEPAConfig, EngineConfig, ...` still raises `ImportError` — the symbols moved to `gepa.legacy_optimize_anything`. Docs/examples were migrated to import from `legacy_*` (good; verified by grep), but external user code using those symbols will still break on merge. |
| 21 | **Issue-level 2026-06-21 — meta_harness missing side_info** | ⚠️ partial | meta_harness *does* surface the info dict — `state/eval_traces/<candidate_name>/*.json` writes the full info dict, and the agent prompt explicitly tells the model to read those files for "compile errors, judge messages, per-example scores, logs, status." So the wiring exists for meta_harness. **Not verified across other engines**: `autoresearch.py` uses the eval.sh subprocess and HTTP response includes `info`, but I didn't trace whether the agent prompt actually reads it. Worth an explicit cross-engine smoke test. |

## Additional issues I found on this pass

These are net-new vs my June 18 review; I've ordered them by criticality.

### C1 — Backwards-compat: `GEPAConfig` import still breaks
**Severity: blocker for external users.** The function call now matches legacy in shape, but every existing user with `from gepa.optimize_anything import GEPAConfig, EngineConfig, ReflectionConfig, MergeConfig, RefinerConfig, TrackingConfig, SideInfo` hits `ImportError`. In-repo docs are migrated to `legacy_optimize_anything` (verified — no broken imports left in `docs/` or `examples/`), but external code wasn't audited and won't be. The simplest fix is **adding a compatibility re-export at the bottom of `optimize_anything.py`**:

```python
# Backwards-compat re-exports for callers using the legacy config classes.
# Will be removed in a future major; users should migrate to OptimizeAnythingConfig.
from gepa.legacy_optimize_anything import (  # noqa: F401, E402
    EngineConfig,
    Evaluator,
    GEPAConfig,
    LogContext,
    MergeConfig,
    OptimizationState,
    RefinerConfig,
    ReflectionConfig,
    SideInfo,
    TrackingConfig,
    get_log_context,
    log,
    make_litellm_lm,
    set_log_context,
)
```

That gets you 100% legacy-import compatibility for zero ongoing maintenance cost.

### C2 — `optimize_anything(..., config=GEPAConfig(...))` silently breaks
A user porting from the legacy API who passes `config=GEPAConfig(...)` (a perfectly natural thing to try) gets an `AttributeError` deep inside `_build_engine` when it accesses `config.max_evals`. There's no `isinstance(config, GEPAConfig)` check that either routes them to `legacy_optimize_anything` or raises a clear `TypeError: GEPAConfig is the legacy config class; use OptimizeAnythingConfig, or import optimize_anything from gepa.legacy_optimize_anything`. Cheap addition; saves an angry forum thread later.

### C3 — Held-out test scoring silently swallows evaluator errors
`_score_test()` at `optimize_anything.py:218–222`:
```python
try:
    score, _ = server.eval_fn(candidate, ex)
    scores[eid] = score
except Exception:
    scores[eid] = 0.0
```
Same shape as the cache-bypass in `engine.py:238` — *bypass* rather than *surface*. A genuine evaluator crash on the test set is indistinguishable from a real 0. The pattern across the new code is "swallow on test, swallow on `process_result`, bypass on cache" — three different places following the same anti-pattern. At minimum, log the exception; ideally also surface counts via `result.metadata["test_errors"]`.

### C4 — `engine.process_result` errors silently swallowed (same pattern)
`optimize_anything.py:175–178`:
```python
try:
    engine.process_result(result, server.output_dir)
except Exception:
    pass
```
A bug in artifact persistence is invisible. Log it at minimum.

### C5 — Cache is still bypassed when `write_agent_state=True`
You flagged this on the previous round (#8 above). Beyond the principled-fix angle, this also means write_agent_state runs do every eval twice for any candidate that gets re-touched — exactly the situation the cache was designed to amortize. If the only thing the cache is missing is *trajectories*, the right fix is to thread `capture_traces` through the cache key (or skip caching just for trace-captures, but log misses), not bypass.

### C6 — `evaluate_examples` HTTP endpoint flattens train + val
`eval_server.py` agent-visible pool is `train_set + val_set` merged for the HTTP layer ("the agent never knows val existed"). For engines that *should* internally treat val differently (e.g., select-on-val, train-on-train), the HTTP layer's flattening loses the distinction. Either expose both, or document the flattening prominently. Combined with your #15 question ("should GEPA even have test?") this is a broader axis the API hasn't picked a stance on.

### C7 — `EvalFn` arity dispatch is fragile
At `eval_server.py:157–160`:
```python
if example is not None:
    score, info = self.eval_fn(candidate, example)
else:
    score, info = self.eval_fn(candidate)
```
The single-task path is now well-defined (`example=None`), so this is better than my earlier "arity detection" criticism. But `(score, info)` still relies on the user returning a 2-tuple — a legacy evaluator returning just `score` (allowed by `legacy_optimize_anything`) will fail with a `TypeError: cannot unpack non-iterable float`. Either normalize (wrap a bare float as `(score, {})`) or document explicitly that the new API requires the 2-tuple.

### C8 — Cross-engine side_info parity (your 06-21 observation)
Your local meta_harness test from June 21 surfaced this. Current state: meta_harness has `state/eval_traces/*.json` written by `_capture_eval_traces` and the agent prompt explicitly instructs reading them. Autoresearch / BestOfN: the `info` dict is in the HTTP `/evaluate` response body, but I couldn't confirm the agent prompt routes the model to consume it. **An explicit cross-engine integration test that asserts each engine's prompt/file layout actually surfaces the side_info dict to the model** is the only way to keep this from regressing as engines are added.

### C9 — `Task.background` "surfaced verbatim by every engine" is still naive
Same as your #18. Particularly painful for subprocess engines where prompt token count is the main cost driver. The natural fix matches the pattern meta_harness already uses for traces: write `background.md` into `work_dir`, instruct the agent to read it. Then the long-form context isn't repeated in every Claude `/print` invocation.

### C10 — `Task.metadata` removed
`8ce195d` removed `Task.metadata: dict[str, Any] = field(default_factory=dict)`. This was an extension point. The replacement is `OptimizeAnythingConfig.engine_config`, which is engine-coupled. Tasks no longer carry domain-typed context (e.g., `task.metadata["run_mode"]`). If this was intentional, fine; if not, it's a silent regression of API surface area. Worth confirming.

### C11 — `optimize_anything_with_server` is in `__all__` but not in the module
Imports happen via `from gepa.oa.ensemble import optimize_anything_with_server`, which works, but the public docs still point at it. Verify that mkdocs-generated pages actually render. (Minor; just spot-check the rendered docs.)

### C12 — Engine `process_result` is "best effort" but its failure mode isn't tested
There's no test that asserts an engine's `process_result` is even called, let alone that artifacts land where expected. Easy to write a fake-engine test: assert the file shows up after a run, and assert a raising `process_result` doesn't crash the run.

### C13 — `Result.best_candidate: str` (was a string-only constraint last review)
Still `str`. The legacy API supported `dict[str, str]` (multi-component candidates) and `str`. Internally the adapter still has the machinery, but `Task.seed_candidate: str | None` and `Result.best_candidate: str` foreclose the multi-component public shape. If a user ports a legacy multi-component setup, they hit a typing wall.

### C14 — Run name uniqueness
`_default_run_name(engine)` returns `{engine}-{uuid8}-{ts}`. The UUID + timestamp combo is plenty unique, but the `output_dir` path also has its own timestamp in `_resolve_output_dir`. Two timestamps a couple lines apart in the same run. Probably OK; mention as nit.

### C15 — `OptimizeAnythingConfig.max_evals = 100` default
`OptimizeAnythingConfig.max_evals: int | None = 100`. That's the same kind of "default too low" that I just landed warnings about in #376 (the budget-floor PR). For any non-trivial valset, 100 evals is well under the ~16 × len(valset) floor. Either:
- Drop the default to `None` and rely on the warning that now fires, or
- Bump to something more like 200 with the same caveat.

Status quo silently configures small runs as a footgun.

## Cross-cutting patterns from your past reviews (#100, dynamic validation)

Things you consistently push for, that this PR could still do better on:

- **"Can we ensure that the end user's experience is not changed."** (PR #100 README comment.) C1 + C2 are the open ones.
- **"We can create a python Protocol for X."** `Engine` is a Protocol — good. `EvalServer` is not. The eval-server interface is now load-bearing for every external engine; promoting it to a `Protocol` (or `ABC`) makes the future "swap in a remote eval server" / "swap in an MCP-backed eval server" story cleaner. Optional, but in the spirit of your PR #100 protocol comment.
- **"Should we introduce a notion of schema versions here?"** (PR #100 state.py comment.) The `iterations/` directory layout and `gepa-result.json` shape are both implicit schemas now. A `_VERSION` constant on the agent-state directory layout would save someone six months from now.
- **DRY in `_save_*` methods** — only partially addressed.
- **Sensible defaults — don't force flags when context implies.** `write_agent_state` is the obvious one. If `run_dir` is set and an agent-proposer is configured, defaulting `write_agent_state=True` is the same logic you've used elsewhere.
- **Naming clarity** — generally improved here (initial→seed, name→config.name); good.

---

## Criticality-ordered fix list

**Blockers (must address before merge):**

1. **`uv.lock` — revert** (your standing comment).
2. **C1 — back-compat re-exports** of `GEPAConfig` / `EngineConfig` / `ReflectionConfig` / `MergeConfig` / `RefinerConfig` / `TrackingConfig` / `SideInfo` / `Evaluator` / `log` / `set_log_context` / `get_log_context` / `make_litellm_lm` from `gepa.optimize_anything`. ~12 lines.
3. **C2 — `isinstance(config, GEPAConfig)` guard** with a clear `TypeError` (or, better, automatic routing to `legacy_optimize_anything.optimize_anything`).
4. **C3 + C4 + C5 — stop swallowing exceptions** in `_score_test`, `process_result` wrapper, and the cache-bypass branch. At minimum `logger.exception(...)`; ideally surface counts.

**High (should address):**

5. **#8 / C5 — extend cache to store trajectories** rather than bypass it. Aligns with your principled-fix comment.
6. **#2 / `write_agent_state` default** — if `run_dir` + agent proposer is configured, default to `True`. The current default forces every agent-proposer user to flip the flag.
7. **C8 — cross-engine `side_info` integration test** so the gap you found locally on meta_harness can't recur silently on a new engine.
8. **#3 / `state.py:38` regex** — exclude `.` (or at least leading `.`).
9. **C9 / #18 — Stop putting `background` in every engine prompt**; write to file for subprocess engines.
10. **C7 — Normalize/document the `(score, info)` contract** so a bare-float evaluator from legacy doesn't blow up.
11. **C15 — `max_evals=100` default** — drop or raise per the budget guide.

**Medium (nice to have, would clean up the API):**

12. **#6 / engine.py:195 consistency question** — author to confirm.
13. **#7 / state.py DRY** — collapse the remaining duplicated save paths.
14. **C13 — re-introduce multi-component candidate shape** (`str | dict[str, str]`) on `Task.seed_candidate` and `Result.best_candidate` if you want to keep that capability publicly.
15. **C10 — confirm `Task.metadata` removal was intentional**; if not, restore as a documented extension point.
16. **C6 — pick a stance on train/val flattening at HTTP** and document.
17. **Schema version stamp** on `iterations/` and `gepa-result.json` (per your #100 pattern).
18. **C5 / autoresearch parity / #5** — at minimum link the reference for the Karpathy autoresearch comparison.
19. **`EvalServer` as a Protocol** for future remote/MCP swaps.

**Low (nits):**

20. **Two timestamps in default output_dir** (C14).
21. **C11 — verify mkdocs renders the API pages**.
22. **C12 — add a test for `process_result` failure isolation**.
23. **Test coverage for the `seed_candidate=None` seedless mode** — I didn't see one explicitly.

Net: the function shape is now genuinely close to legacy, which is the right move. The remaining backwards-compat gap (C1 + C2) plus the swallowed-exceptions pattern (C3 + C4 + C5) are the two patterns I'd want resolved before this lands.

tests/test_attach_existing_run.py
```diff
@@ -8,19 +8,19 @@
 import pytest
 
 from gepa.logging.experiment_tracker import ExperimentTracker, create_experiment_tracker
-from gepa.optimize_anything import GEPAConfig, TrackingConfig
+from gepa.legacy_optimize_anything import GEPAConfig, TrackingConfig
```
LakshyAAAgrawal: Almost ALL test changes, in files that were earlier importing gepa.optimize_anything are now instead testing legacy_optimize_anything.

Shouldn't the tests actually be targetting the new API? Most of the tests should be applicable to the new API as well.

Shangyint: not really. Now legacy_optimize_anything becomes one engine of optimize_anything (as a gepa wrapper). Therefore, lots of tests are not applicable to `optimize_anything`. If we desire any of these features, we have to implement it for all engines (and first add it to the engine class contract), then we can support the feature generally for `optimize_anything`.

LakshyAAAgrawal: GEPA was always supposed to be just one of optimize_anything engines. Can we migrate all tests to the new API. For the tests that are GEPA specific, we should run it through the new api with engine='gepa'. Ideally, we would want to deprecate legacy_opt_any over the next 2 releases, so all tests should exercise the new default path.

Shangyint: Some of the test do not just use legacy_opt_any, it imports and tests other configs and helper functions from the file like 

```
    EngineConfig,
    GEPAConfig,
    ReflectionConfig,
    _build_seed_generation_prompt,
    _generate_seed_candidate,
```

IMO we should probably rename legacy_opt_any to something else. We cannot really deprecate the whole file.

What I agree with: any call to legacy_opt_any should route through the new api for sure.

LakshyAAAgrawal:
### `optimize_anything` orphans `GEPAConfig.tracking` — native wandb/mlflow is silently disabled (plus a few other core features)

Running the `gepa` engine through `optimize_anything` produces a near-empty wandb dashboard — only
`eval/score`, `eval/best_score`, `eval/used`, and per-iteration metrics. None of GEPA core's rich
tracking (Pareto-front evolution, per-candidate tables, the candidate-tree HTML) shows up, even
though `docs/docs/guides/experiment-tracking.md` advertises it and the standalone `gepa.optimize()`
path produces it.

**Root cause.** `core` exposes tracking as a top-level `GEPAConfig` field:

```python
# gepa_launcher.py
@dataclass
class GEPAConfig:
    engine: EngineConfig
    reflection: ReflectionConfig
    tracking: TrackingConfig      # use_wandb / wandb_attach_existing / wandb_step_metric / mlflow
    merge: MergeConfig | None
    refiner: RefinerConfig | None
    stop_callbacks: ... 
    callbacks: ...
```

But `GepaEngine` (`oa/engines/gepa.py`) only forwards a subset of `OptimizeAnythingConfig.engine_config`:

```python
_GEPA_CONFIG_KEYS = ("engine", "reflection", "merge", "refiner",
                     "objective", "background", "reflection_lm_kwargs",
                     "callbacks", "claude_code_agent")
# no "tracking" → GEPAConfig(...) is built without a TrackingConfig
```

So GEPA core runs with its default (stdout) tracker. The only wandb signal is the **separate, thinner**
`OptimizeAnythingConfig.tracker` hook (`log_eval(used, score, best_score, cost)` / `log_metrics`),
fired from the eval server and `_ProgressCallback`. Worse, because `tracking` isn't in
`_GEPA_CONFIG_KEYS`, a user who *tries* `engine_config={"tracking": {"use_wandb": True}}` gets it
**warned-and-dropped** — there's no way to turn the native tracker on through the OA path.

This is unfortunate because the native `TrackingConfig` is excellent and seems purpose-built for this
exact embedding: `wandb_attach_existing=True` + `wandb_step_metric="gepa/iteration"` would log GEPA's
full dashboard **into the caller's existing run** without step-collisions. `mlflow` is orphaned the
same way.

**Suggested fix.** Have `GepaEngine` translate tracking config into the `GEPAConfig` it builds — e.g.
add `"tracking"` to `_GEPA_CONFIG_KEYS` and pass `GEPAConfig(tracking=TrackingConfig(**tracking), ...)`,
and/or, when `OptimizeAnythingConfig.tracker`/a wandb run is active, default to
`TrackingConfig(use_wandb=True, wandb_attach_existing=True, wandb_step_metric="gepa/iteration")`. Then
document, in `experiment-tracking.md`, that the OA `tracker` field and core `TrackingConfig` are two
different things.

---

**Same mechanism hides other core features behind the OA layer.** Because `GepaEngine` only forwards
the keys above, anything that lives at `GEPAConfig` *top level* (rather than inside the four
sub-config dicts) is unreachable via `optimize_anything`:

- **`tracking`** — wandb + mlflow (above).
- **`stop_callbacks`** — the OA path hardcodes only `ScoreThresholdStopper` (from `stop_at_score`)
  and `MaxReflectionCostStopper` (from `max_token_cost`). The other exported stoppers
  (`NoImprovementStopper`, `MaxMetricCallsStopper`, `FileStopper`, `SignalStopper`,
  `CompositeStopper`) and any custom `StopperProtocol` can't be passed through.
- **Multi-component candidates** — core `optimize_anything` takes `seed_candidate: str | Candidate`
  where `Candidate = dict[str, str]` (jointly optimizing several named components, with the evaluator
  receiving the dict). The OA entry point and `Task.seed_candidate` are **`str` only**, so
  multi-component / multi-module optimization — one of GEPA's signature capabilities — isn't reachable
  through `optimize_anything`.
- **Custom `logger` (`LoggerProtocol`)** — the lower-level core entry point accepts a `logger`; the
  OA path doesn't expose one.

**And a set of features that *are* reachable but effectively undiscoverable** — they only work via the
opaque `engine_config["engine"|"reflection"]` kwargs passthrough, aren't surfaced on
`OptimizeAnythingConfig`, and aren't shown in any OA example:

- All `EngineConfig` knobs, including **`seed` (reproducibility)**, `frontier_type`
  (`"hybrid"`/`"objective"`/...), `candidate_selection_strategy`, `acceptance_criterion`,
  `val_evaluation_policy`, `num_parallel_proposals`, `display_progress_bar`, `track_best_outputs`,
  `best_example_evals_k`.
- **Multi-objective Pareto.** The adapter *does* honor `info["scores"]` → `objective_scores`
  (`optimize_anything_adapter.py`) and `frontier_type` defaults to `"hybrid"`, so you can optimize,
  e.g., correctness *and* speedup jointly — but the `info["scores"]` contract is documented nowhere a
  user would find it; the CUDA-kernel appendix returns a scalar score and never mentions it.

Net: the engine-as-narrow-passthrough design means GEPA's `tracking`, `stop_callbacks`,
multi-component candidates, and `logger` are silently lost through `optimize_anything`, and a large
slice of `EngineConfig`/multi-objective behavior is reachable only by reading core's dataclasses. A
short "GEPA via optimize_anything: what maps to what" doc plus forwarding `tracking`/`stop_callbacks`
(and a string-or-dict `seed_candidate`) would close most of the gap.

LakshyAAAgrawal:
The OA tracker is untyped and its contract is undocumented

OptimizeAnythingConfig.tracker: Any with the docstring "wandb/mlflow wrapper", but the required
methods — log_eval(used, score, best_score, cost) and log_metrics(metrics, step=None) — are
discoverable only by grepping oa/eval_server.py and the callbacks. Given Engine got a proper
Protocol, the tracker deserves one too (e.g. an OATracker Protocol), and
guides/experiment-tracking.md should document the OA tracker explicitly and contrast it with the
core ExperimentTracker (they are different objects with different method sets — a real source of
confusion, see #2).

engine_config typo handling. Unknown keys warn (good, and documented) but warnings are
easy to miss in long runs. Consider an opt-in strict mode, and recording the accepted-vs-ignored
keys in the run summary so a silent typo is auditable after the fact.

Two same-named entry points. gepa.optimize_anything.optimize_anything
(OptimizeAnythingConfig) vs gepa.gepa_launcher.optimize_anything (GEPAConfig) share a name
and most of a signature but take different config classes.

With the introduction of the new API, all the code examples in the optimize_anything blogpost will break, but it should be cleanly migrated to the new API.

Shangyint:
Yeah agreed. Logger/tracker seem to have issues. I’ll add tests and fix
bugs.

Shangyint:
@LakshyAAAgrawal latest commit refactored a bunch of things: now `engine_config` is parsed exactly into our old `GEPAConfig` (in the original `optimize_anything`). Now the tracking should work exactly the same as the old one! For autoresearch or mh tracking, we should address them in a later PR.

LakshyAAAgrawal:
Hey @Shangyint, I just merged #369, which touches many of the files this PR did too. Can you please resolve conflicts and rebase on top of main?

Shangyint:
@LakshyAAAgrawal yep merged. The conflicts seems additive, which is nice.

PR #374, opened 2026-06-11
feat: Add Trackio Support / total_metric_calls on callback events

evalstate:
# Summary

Adds Trackio alongside WandB and MLflow.

- Supports metrics, config, summaries, tables, artifacts, key prefixes, custom step metrics, and existing-run attachment.
- Keeps Trackio optional (`pip install trackio`).
- Adds committed GEPA metric-call accounting to `ValsetEvaluatedEvent`.
- Fixes iteration-0 valset events to include seed outputs when `track_best_outputs=True`.

## Trackio Integration 

This adds [`Trackio`](https://github.com/gradio-app/trackio) support to GEPA for metric logging and reporting. 

It follows the existing patterns for metrics integrations, and includes relevant document updates.

I've been extensively testing Trackio + GEPA recently.

## `total_metric_calls` on Callback Events

(Happy to split this in to a separate PR if desired) 

When wiring this up I found that callback consumers can't plot against the
cumulative metric-call budget without re-deriving it from their own bookkeeping. This
adds a `total_metric_calls` field to `EvaluationEndEvent`, `EvaluationSkippedEvent`,
`ProposalEndEvent` and `ValsetEvaluatedEvent`.

The value is the cumulative budget consumed as of the end of the work each event
describes. Events emitted before `state.increment_evals()` therefore include the
in-flight count for the evaluation being reported, so the number always agrees with
the event's own payload. The field is `NotRequired`, so any third-party emitters of
these events keep type-checking.

This also fixes the iteration-0 `ValsetEvaluatedEvent`, which passed
`outputs_by_val_id=None` unconditionally while every later iteration forwards the
evaluation's outputs — consumers saw a hole at iteration 0 only.

Best, evalstate.

---


uv.lock
LakshyAAAgrawal, line 1: Please revert changes to this file

pyproject.toml
LakshyAAAgrawal, line 1: Please revert this change as well. We can expect the user will install experiment trackers separately. If trackio is enabled, there should be a:
```
try:
    import trackio
except:
    print("Install trackio by pip instal ...")

evalstate:
Thanks for taking a look at this @LakshyAAAgrawal; I've taken trackio out of the project deps as advised.

evalstate:
end-to-end simple GEPA run
image

edited with more useful image.

fast-agent throughput/llm metrics
image

Issue #251, opened 2026-03-13
Bandit-learning based proposal strategy selection


LakshyAAAgrawal:
## Summary

Currently, GEPA uses a fixed `proposal_mode` (rewrite, edit, or crossover) for the entire optimization run. Drawing inspiration from ShinkaEvolve's stochastic patch type selection and adaptive LLM model selection, we should implement a bandit-learning based mechanism that dynamically selects the best proposal strategy per iteration.

## Motivation

Different proposal strategies work better in different contexts:
- **Rewrite** is best when candidates need major structural changes
- **Edit** (search/replace) is best for long candidates where only small sections need tuning — saves tokens and preserves working sections
- **Crossover** is best when the population has diverse candidates with complementary strengths

A fixed strategy forces users to guess which will work best. An adaptive mechanism can learn from optimization history which strategy produces the most improvement, and shift probability mass accordingly.

### `proposal_mode="auto"` (bandit selection)

When `proposal_mode="auto"`, each iteration samples a proposal strategy from a learned probability distribution.

**Bandit algorithms to support:**
- **Thompson Sampling** (default) — maintains Beta distributions over success probability for each strategy; naturally balances exploration/exploitation
- **UCB1** — upper confidence bound; deterministic, good for lower variance
- **Epsilon-greedy** — simple baseline; exploit best strategy with probability 1-ε, explore uniformly with probability ε

**Reward signal:** Whether the proposed candidate improves on the parent's subsample score. Binary reward (improved vs. didn't improve) feeds into the bandit.

**Configurable prior probabilities:** Allow users to set initial weights, e.g., `proposal_strategy_weights={"rewrite": 0.3, "edit": 0.6, "crossover": 0.1}` matching ShinkaEvolve's defaults.

### Implementation outline

1. Add `ProposalStrategySelector` class in `src/gepa/strategies/` with pluggable bandit algorithms
2. Track per-strategy success/failure counts in `GEPAState` (so it persists across checkpoint/resume)
3. In `ReflectiveMutationProposer.propose()`, when `proposal_mode="auto"`, use the selector to pick the strategy for this iteration
4. After evaluating the proposed candidate, update the bandit with the reward signal
5. Log the selected strategy and bandit state to experiment tracker

### References

- ShinkaEvolve uses fixed stochastic sampling: `patch_types=["diff", "full", "cross"]`, `patch_type_probs=[0.6, 0.3, 0.1]`
[...]

PR #389, opened 2026-07-11
feat(proposer): add CrossoverProposer for LLM-synthesized recombination

jiwoomap:
## Summary

Adds an opt-in `crossover` strategy that recombines two Pareto candidates with
complementary strengths by asking the reflection LM to synthesize a single component
text that integrates both parents. This fills a gap between the two existing
proposers: reflective mutation edits one candidate at a time, and `MergeProposer`
recombines only along a shared genealogy, by swapping whole component texts rather
than synthesizing them.

## Why this helps

Prompt optimization can get stuck with candidates that are each strong on a
different slice of the data. For example:

> Candidate A: strong on structured/format-heavy inputs (its instruction spells out
> output formatting rules) but weak on open-ended reasoning.
> Candidate B: strong on open-ended reasoning (its instruction emphasizes
> step-by-step decomposition) but sloppy on formatting.

GEPA currently has no dedicated mechanism to combine them. Crossover asks the
reflection LM to integrate A's and B's strengths into a single instruction, using
evidence of where each parent wins.

Concretely, this can provide:

1. A higher achievable ceiling under the same rollout budget, by combining complementary strengths instead of waiting for a single mutation to rediscover both.
2. A way to escape stagnation when candidates specialize in different subsets of the data.
3. Limited additional rollout cost, since synthesis uses the reflection LM and only the child evaluation consumes metric calls.

This only pays off when the population actually contains candidates with different
strengths, which is what the Pareto frontier tends to produce.

## Relationship to #252 / #251

#252 ("Add LLM-based crossover proposal mode") explores the same intuition — asking the
LLM to combine two candidates — but at a different layer, and I believe the two are
complementary rather than competing:

| | #252 | this PR |
|---|---|---|
| Layer | a `proposal_mode` of reflective mutation (applies to every mutation for the whole run) | a scheduled proposer slot alongside merge (fires only after an accepted iteration) |
| Partner selection | random donor from the population | pair of Pareto candidates chosen by per-instance win-set complementarity |
| LM input | the two texts | the two texts + contrastive evidence of where each parent wins |
| Acceptance | same as any mutation | contested-subsample gate: `sum(after) >= max(parents)` before the full valset eval |

If the bandit-based strategy selection proposed in #251 lands, this proposer could be
registered as one of its arms alongside rewrite/edit. Happy to align naming or
consolidate however the maintainers prefer.

## What changed

- `src/gepa/proposer/crossover.py` (new): `CrossoverSignature` (reflection prompt
  integrating two parent texts given evidence of where each wins) and
  `CrossoverProposer`:
  - Parent selection by complementarity, not genealogy — picks the pair of Pareto
    candidates whose per-instance win-sets are most mutually substantial
    (ranked by `min(|A wins|, |B wins|)`, gated by `min_win_asymmetry`).
  - Synthesis, not swap — recombines one divergent component via a single
    reflection-LM call.
  - Synthesis runs on the reflection LM; the only metric calls are the child
    subsample eval, plus the full valset eval on acceptance.
  - Retries a pair on other divergent components and retires it once exhausted.
  - Emits `on_evaluation_start` / `on_evaluation_end` around the contested-subsample
    eval, matching `MergeProposer`.
- `src/gepa/core/callbacks.py` — crossover telemetry mirroring merge's:
  `CrossoverAttemptedEvent` / `CrossoverAcceptedEvent` / `CrossoverRejectedEvent`
  TypedDicts, the corresponding `GEPACallback` protocol methods, and
  `CompositeCallback` forwarding.
- `src/gepa/core/engine.py` — a crossover slot mirroring the merge slot
  (scheduled after a reflective accept; accepts when `sum(after) >= max(parents)`).
  The engine emits `on_crossover_attempted` before the acceptance check and
  `on_crossover_accepted` / `on_crossover_rejected` on each outcome. New engine arg
  `crossover_proposer` defaults to `None`, so existing callers are unaffected.
- `src/gepa/api.py` — new flags: `use_crossover` (default `False`),
  `max_crossover_invocations` (5), `crossover_min_win_asymmetry` (2).
  `use_crossover=True` requires a `reflection_lm`.
- `docs/` — `CrossoverProposer` and the three crossover events registered in the
  API docs and nav (via `generate_api_docs.py`), plus a crossover section in the
  callbacks guide.

## Behavior / compatibility

- Default off; existing runs are unchanged.
- When both merge and crossover are enabled, merge takes priority within an
  iteration (its slot runs first and `continue`s on an attempt).

## Testing

- `tests/proposer/test_crossover.py` — unit coverage for pair selection (including
  valid-but-lopsided pairs and multi-component retry), synthesis, skip conditions,
  subsample balancing, and input validation.
- `tests/proposer/test_crossover_integration.py` — end-to-end acceptance and
  rejection paths, plus the `use_crossover` without `reflection_lm` error.
- `tests/test_callbacks.py` — `TestCrossoverEvents` mirroring `TestMergeEvents`;
  `RecordingCallback` in unit and integration callback tests extended with the new
  protocol methods.
- Full suite: 581 passed, 4 skipped (after rebasing onto v0.1.4). `ruff` + `pyright`
  clean on changed files.

## Notes for reviewers

- `max_crossover_invocations` caps accepted crossovers (mirroring
  `max_merge_invocations`); attempts are separately bounded by the attempted/
  exhausted-pair bookkeeping.
- Known minor follow-up (not blocking): guard against a synthesis that echoes a
  parent verbatim (would add a duplicate candidate).


jiwoomap:
## Empirical demo: crossover firing with a real LLM

To sanity-check the design beyond the mocked tests, I ran a small live experiment with `gemini-2.5-flash` as both task LM and reflection LM.

**Setup.** 8 Korean office-arithmetic questions, two hidden output rules that act as separate skill axes: calc questions must end with `### <digits>`, time questions with `@@@ HH:MM`. Feedback only reveals the rule of the failed question's own type, and `reflection_minibatch_size=2` means a single mutation usually learns one rule — deliberately creating the "two half-strong candidates" regime crossover targets.

**Result** (`use_crossover=True`, `max_metric_calls=140`, ~3 min):

| idx | parents | total | calc axis | time axis | how |
|---|---|---|---|---|---|
| 0 | – | 0.26 | 0.30 | 0.22 | seed |
| 1 | [0] | 0.50 | **0.00** | **1.00** | mutation — time specialist, calc regressed |
| 2 | **[0, 1]** | 0.65 | 0.30 | 1.00 | **crossover** — kept [1]'s time mastery, restored [0]'s calc behavior |
| 3 | [2] | **1.00** | 1.00 | 1.00 | mutation on the fused base |

Notes:
- The pair (0, 1) was selected automatically from per-instance win-sets — [0] uniquely won the 4 calc instances, [1] the 4 time instances.
- Candidate [1] was a local-optimum trap (highest total at the time, but calc broken at 0.00); the crossover child escaped it by re-injecting the seed's calc behavior without regressing the time axis, and one further mutation finished the job.
- On an easier variant of the task (single rule), the first mutation reached 1.00 and the crossover slot correctly never fired — confirming it stays out of the way when there is nothing to fuse.

The same accept/reject mechanics are covered deterministically in `tests/proposer/test_crossover_integration.py`; this was an additional live sanity check.

jiwoomap:
@LakshyAAAgrawal Gentle ping on this one
it's been open since July and I'd love to get your take on whether the direction fits.

PR #391, opened 2026-07-14
fix: cap litellm <1.92 on Python 3.14 and make CI dependency resolution deterministic

LakshyAAAgrawal:
## What broke

The push-to-main run after the v0.1.2 release merge failed on the Python 3.14 jobs only ([Run Tests (3.14)](https://github.com/gepa-ai/gepa/actions/runs/29331310678/job/87079651699), [Build Package (3.14)](https://github.com/gepa-ai/gepa/actions/runs/29331310678/job/87079651729); 3.10/3.13 were fail-fast-cancelled, 3.11/3.12 passed).

**Root cause has two layers:**

1. **Upstream:** litellm 1.92.0 (released 2026-07-12) is the first litellm release with a native extension (`litellm-rust`, maturin/PyO3 0.23.5). Its wheels are **manylinux-only, cp310–cp313** — no macOS, no Windows, no cp314. Everyone outside that matrix falls back to the sdist, which requires Cargo ≥ 1.85 (`edition2024`) and a PyO3 that caps at Python 3.13. CI only surfaced the 3.14 slice because runners are Linux — **fresh installs of `gepa[full]` are also broken for macOS and Windows users on every interpreter** (reproduced locally on macOS arm64 / Python 3.12). Every litellm ≤ 1.91.x was a pure-Python wheel, which is why this never happened before.
2. **Our CI:** `uv.lock` pins litellm, but the release automation bumps the version in `pyproject.toml` **without refreshing `uv.lock`**, making the lock stale — and plain `uv sync` then silently re-resolves against latest PyPI instead of failing. That's how a 2-day-old upstream release reached a run triggered by a version-bump-only commit. **Not a code regression** in the release merge.

## Changes

- **`pyproject.toml`** — `litellm>=1.83.0,<1.92` **unconditionally** (`full` and `confidence` extras), with a comment on exactly what to check before lifting (complete wheel matrix or a pure-Python build). An earlier revision scoped the cap to `python_version >= '3.14'`; review scrutiny + local reproduction showed the platform dimension makes that too narrow.
- **`uv.lock`** — re-locked: records 0.1.2 (un-stranding the lock), litellm at 1.91.3 (pure-Python) everywhere, and wandb re-resolved off the yanked 0.24.0 (→ 0.28.0).
- **`run_tests.yml`** — `uv sync --locked`: a stale lock now fails loudly instead of silently re-resolving, so dependency changes must ship the updated lock in the same PR, where CI tests exactly what's locked. Plus `fail-fast: false` so one interpreter's failure doesn't cancel the rest of the matrix signal.
- **`build_and_release.yml`** — the version-bump step now runs `uv lock` and commits it with `pyproject.toml`, closing the stale-lock loophole that triggered this incident.
- **`dependency_canary.yml`** (new) — scheduled (Mon/Thu) + manual. Uses **`uv pip install` (per-interpreter, pip-semantics resolution)**, not `uv lock --upgrade`: a universal lockfile unifies a package to one version whenever a single version satisfies every marker branch, so re-locking cannot simulate what real users resolve (verified: `uv pip install` picks litellm 1.92.0 on 3.12 where the universal lock pins 1.91.3). Matrix spans **ubuntu + macos × 3.10–3.14** since wheel gaps are platform- and ABI-specific; failures open/update a tracking issue instead of sitting silently in the Actions tab.

## Trade-off, stated explicitly

PR CI (`--locked`) is deterministic and therefore tests pinned versions, not latest — e.g. litellm 1.91.3 on all interpreters. That's by design: determinism in PRs, with the canary as the compensating control that tests fresh per-interpreter resolution on a schedule. Before this PR, CI was neither deterministic nor latest — it depended on whether the lock happened to be stale.

## Verification

- `uv lock --check` passes (the new CI gate).
- Locked path (what PR CI runs), Python 3.12: **476 passed, 4 skipped** (pre-existing unrelated skips).
- Fresh pip-semantics install on macOS arm64 / 3.12 — the case that fails on current `main` metadata — resolves litellm 1.91.3 and passes the full suite: **476 passed, 4 skipped**.
- This PR's own CI matrix validates the five Linux interpreters including both previously-red 3.14 jobs.

## Follow-ups (not in this PR)

- Cut **v0.1.3** after merge so PyPI users (macOS/Windows all Pythons; Linux 3.14) get the fixed metadata.
- Lift the `<1.92` cap once litellm ships a complete wheel matrix (tracked in the pyproject comment).

### DIFF
--- a/.github/workflows/build_and_release.yml
+++ b/.github/workflows/build_and_release.yml
@@ -62,8 +62,14 @@ jobs:
           NEW_VERSION=$(uv run python .github/workflows/build_utils/test_version.py $PACKAGE_NAME $VERSION https://test.pypi.org)  
           echo "Version to be used for TestPyPI release: $NEW_VERSION"  
           echo "version=$NEW_VERSION" >> "$GITHUB_OUTPUT"
-      - name: Update version in pyproject.toml
-        run: sed -i '/#replace_package_version_marker/{n;s/version="[^"]*"/version="${{ steps.check_version.outputs.version }}"/;}' pyproject.toml
+      - name: Update version in pyproject.toml and refresh lockfile
+        run: |
+          sed -i '/#replace_package_version_marker/{n;s/version="[^"]*"/version="${{ steps.check_version.outputs.version }}"/;}' pyproject.toml
+          # Keep uv.lock in sync with the bumped version. A stale lock breaks
+          # `uv sync --locked` in CI, and with a plain `uv sync` it silently
+          # re-resolves against latest PyPI (which is how an unbuildable
+          # litellm 1.92.0 reached the Python 3.14 jobs after the 0.1.2 bump).
+          uv lock
       - name: Build a binary wheel
         run: uv run python -m build
       - name: List built distributions
@@ -91,7 +97,7 @@ jobs:
       - name: Create release-test branch and push changes
         run: |
           git checkout -b release-test-${{ needs.extract-tag.outputs.version }}
-          git add pyproject.toml
+          git add pyproject.toml uv.lock
           git commit -m "Update version to ${{ steps.check_version.outputs.version }} for test PyPI (from tag ${{ needs.extract-tag.outputs.version }})"
           git push origin release-test-${{ needs.extract-tag.outputs.version }}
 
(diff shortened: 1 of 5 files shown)


LakshyAAAgrawal:
Re: review — both findings verified and addressed, and the first one led somewhere bigger. Thank you.

**1. Unification finding — confirmed, then overtaken by a broader fix.** Confirmed empirically: `uv lock --upgrade` produces a single `[[package]]` litellm 1.91.3 entry with no resolution-marker fork (contrast wandb, which genuinely forks), exactly as you described. While rehearsing the canary locally to double-check the consequence, a fresh pip-semantics install on **macOS arm64 / Python 3.12** failed: litellm 1.92.0's wheels turn out to be **manylinux-only** (no macOS/Windows wheels at all, not just no cp314), and the sdist needs Cargo ≥ 1.85. So the `>= '3.14'` marker was too narrow on the platform axis, not just imprecise on the version axis — the cap is now **unconditional**, which also makes metadata and lock agree on `<1.92` everywhere by design rather than by resolver accident.

**2. Canary blind spot — fixed structurally.** The canary now uses `uv pip install` (per-interpreter, pip-semantics, latest-matching) instead of `uv lock --upgrade`, since a universal lock structurally can't simulate user resolution (verified: `uv pip install` picks 1.92.0 on 3.12 where the lock pins 1.91.3). Matrix widened to **ubuntu + macos × all five interpreters** — this incident is itself the proof that platform-specific wheel gaps are a real class. Failure now opens/updates a tracking issue (`issues: write` added).

**3. Other notes:** release-job duplication acknowledged — left as-is at 2×, composite action if that file grows. Drive-by while re-locking: wandb was pinned at the now-yanked 0.24.0; re-resolved to 0.28.0.

Verified locally: `uv lock --check` OK; locked path 476 passed / 4 skipped; the previously-failing macOS fresh install now resolves 1.91.3 and passes the full suite.

LakshyAAAgrawal:
Addressed the duplicate-issue race as well, since the fix was structural rather than cosmetic: notification moved out of the matrix legs into a single `notify-failure` job that runs once after all legs settle (`needs: fresh-resolve`, `if: needs.fresh-resolve.result == 'failure'`). No concurrent `gh issue create` is possible by construction, the issue body now names every failed leg via the jobs API, and `issues: write` is scoped down from workflow level to just the notifier job.

PR #369, opened 2026-06-04
feat: ReflectionLM protocol + batch-based parallel proposals (#329 Phases 1 & 4)

Benzhang2004:
## Summary

Implements **Phase 1** and **Phase 4** of the unified reflection LM design (#329), on top of `main`. Two focused commits:

### Phase 1 — ReflectionLM protocol + StatelessReflectionLM
- New `proposer/reflective_mutation/reflection_lm.py`: a `ReflectionLM` whose `reflect()` returns `(proposal, next_lm)` — the stateful-capable signature from #329. The stateless default (`StatelessReflectionLM`) returns itself, so behavior is identical to today.
- `ReflectiveMutationProposer.propose_new_texts` now delegates to it; the per-component reflection loop and missing-template warning moved into `StatelessReflectionLM`.

### Phase 4 — batch-based parallel proposals (replaces #314)
- Reverts #314's `ThreadPoolExecutor` parallelism. Parallelism now lives at the edges — `adapter.batch_evaluate()` and the LM — keeping GEPA's engine sequential and async-ready (no threading in core state management).
- **Sampling strategies** (`strategies/proposal_sampling.py`): `SingleMutationSampling` (default), `SameParentSampling(n)`, `IndependentSampling(n)`, `PxNSampling(p, n)`.
- **Selection strategies** (`strategies/proposal_selection.py`): `AllImprovements` (default), `BestImprovement`, `TopKImprovements(k)`.
- `ReflectiveMutationProposer.propose()` is a 5-stage pipeline returning `list[CandidateProposal]`: sample → batch-eval parents (deduplicated) → reflect per task (via the Phase 1 `ReflectionLM`) → batch-eval children → select.
- Optional `batch_evaluate` on the `GEPAAdapter` protocol with a `default_batch_evaluate()` sequential fallback; `batch_evaluator` wired through `optimize_anything()` → `OptimizeAnythingAdapter` (incorporates #279).
- Engine has **one** code path: always calls `propose()`, iterates results. Removes `num_parallel_proposals`, `ProposalContext`/`ProposalOutput`, and the threaded batch runner.

**Defaults preserve behavior exactly:** `SingleMutationSampling` + `AllImprovements` reproduce the original single-proposal flow.

## Deferred

Additional reflection-LM / proposer variants from #329 are **not** in this PR and are deferred to follow-ups:
- **ComBEE** map-shuffle-reduce reflection (#307) — Phase 3.
[...]

Issue #329, opened 2026-04-11
Design: Unified Reflection LM Pool — sessions, agents, and ComBEE as first-class reflection LMs


LakshyAAAgrawal:
## Motivation

GEPA currently has a single `reflection_lm` (a stateless callable) that proposes new candidates each iteration. Several in-flight PRs need richer reflection primitives:

- **Sessions (#305):** Stateful LLM conversations that carry context across iterations, with fork/reset lifecycle
- **Coding agents (#293, #305):** CLI tools (Claude Code, OpenCode) that edit files and maintain session state
- **ComBEE (#307):** Map-shuffle-reduce aggregation of reflections across large minibatches
- **Parallel proposals (#314):** Multiple reflection calls per iteration from different parents

Each PR introduces its own abstractions (`SessionManager`, `SessionStrategy`, `CodingAgent`, `ComBEEConfig`, `ParallelMutationOrchestrator`). This proposal unifies them under a single concept: **a pool of reflection LMs**.

### 1. A reflection LM returns both a new candidate AND a new (extended) reflection LM

```python
class ReflectionLM(Protocol):
    def reflect(
        self,
        candidate: dict[str, str],
        feedback: Mapping[str, Sequence[Mapping[str, Any]]],
        components_to_update: list[str],
    ) -> tuple[dict[str, str], "ReflectionLM"]:
        """Propose new component texts and return an extended copy of this LM.

        Returns:
            (new_texts, extended_lm) where:
            - new_texts: dict mapping component name -> proposed new text
            - extended_lm: a new ReflectionLM with this call's context appended.
              The original LM is unchanged (can be reused/forked).
        """
        ...
```

The returned `extended_lm` is the original session continued — it has seen this candidate and feedback. The original is immutable (safe to reuse from the pool).

### 2. GEPA maintains a pool of reflection LMs

Just as GEPA maintains a pool of candidate programs, it maintains a **pool of reflection LMs**. Each iteration, a selection strategy picks both:
- A **candidate** from the candidate pool (existing `CandidateSelector`)
- A **reflection LM** from the reflection LM pool (new `ReflectionLMSelector`)

After the call, the extended LM goes back into the pool. The pool always contains at least one **empty LM** (no prior context) — selecting it is equivalent to today's fully stateless behavior.

### 3. This unifies sessions, agents, and ComBEE

| Concept | How it maps to a reflection LM |
|---|---|
[...]

PR #307, opened 2026-04-03
ComBEE parallel scan aggregation for scalable prompt learning

nuglifeleoji:
Hi, this is a pr that implements this paper's algorithm on gepa, here are the overall summary and usage, feel free to ask questions and comment!

## Summary

Adds `aggregation_method` parameter to support **ComBEE** (Combee: Scaling Prompt Learning for Self-Improving Language Model Agents), a Map-Shuffle-Reduce framework that prevents context overload when using large reflection batch sizes.

When `aggregation_method="naive"` (default), the existing code path is completely unchanged. When set to `"combee"`, the reflection step applies:

- **Augmented shuffling (§3.2)** — each reflection is duplicated `p=2` times and the full set is shuffled, ensuring every reflection has multiple chances to be incorporated
- **Parallel scan aggregation (§3.1)** — reflections are split into `k = ⌊√n⌋` groups; Level-1 calls one reflection LM per group to produce `k` intermediate instructions; Level-2 aggregates those `k` instructions into a single final context update

This resolves the context overload problem where naive large-batch reflection degrades quality because the aggregator LLM defaults to retaining only generic patterns while discarding specific, high-value insights.

## Usage

```python
# gepa.optimize
result = gepa.optimize(
    ...,
    reflection_minibatch_size=40,   # this is n in the paper
    aggregation_method="combee",    # k = ⌊√40⌋ = 6 groups computed automatically
)

# optimize_anything
config = GEPAConfig(
    reflection=ReflectionConfig(
        reflection_minibatch_size=40,
        aggregation_method="combee",
    )
)
```

`aggregation_method="naive"` is the default — no behaviour change for existing users.

## Implementation

- **`src/gepa/strategies/instruction_proposal.py`** — added `default_aggregation_prompt_template` to `InstructionProposalSignature`, the Level-2 reduce prompt that synthesizes `k` intermediate instructions into one
- **`src/gepa/proposer/reflective_mutation/reflective_mutation.py`** — added `_propose_new_texts_combee()` implementing the full Map-Shuffle-Reduce logic; `propose()` routes to it when `aggregation_method="combee"` and `reflection_lm` is set (falls back to naive otherwise)
- **`src/gepa/api.py`** — `aggregation_method` exposed on `gepa.optimize()`
- **`src/gepa/optimize_anything.py`** — `aggregation_method` added to `ReflectionConfig` with docstring

ComBEE is only active when `reflection_lm` is provided and no custom adapter proposer overrides the default LM-based reflection.

PR #392, opened 2026-07-14
Review follow-ups for #369: child-eval events, recorded-run behavior pins, reflection_strategy seam, docs

LakshyAAAgrawal:
Stacked on #369 (base: `feat/reflection-lm-protocol-phase1`). Three focused commits addressing review findings:

## 1. `fix`: restore `on_evaluation_start`/`on_evaluation_end` for child candidate evals
The batch `propose()` pipeline emitted evaluation events only around the **parent** minibatch evaluation; the pre-batch sequential path on `main` emits them at both the parent and the child (new candidate) sites. This restores the child-side events with the same payload shape as `main` (`candidate_idx=None`, `parent_ids=[parent_idx]`) — same family as the `on_evaluation_skipped` regression already fixed in `f8b7f56e`.

Also replaces the `hasattr`-only `TestDefaultStrategiesRetainBehavior` with real behavior pins against the recorded AIME replay run (`tests/test_aime_prompt_optimization/`). Because the replay fixture hard-fails on any unseen LM input, a passing run proves the default `SingleMutationSampling`+`AllImprovements` path reproduces the recorded (pre-#369) run's exact LM request stream. New pins:
- default path reproduces the golden optimized prompt
- budget-ledger consistency: monotonic, self-consistent deltas; only the seed valset eval precedes budget-hook registration; final == `result.total_metric_calls`
- explicit default strategies are event-for-event identical to implicit defaults
- proposal event ordering
- pinned budget-event composition (2 per completed proposal + 1 per accepted candidate) and child-evaluation event parity — so any future change to event granularity is made consciously

**On budget-event granularity:** iterations now emit separate parent-eval and child-eval `on_budget_updated` events instead of one combined event. Totals are unchanged (verified by the pins above); no in-repo consumer counts events (trackers don't subscribe; stoppers/progress read `state.total_num_evals` directly); and per-stage deltas are required once sampling strategies produce n>1 proposals. Documented in the new guide rather than reverted.

## 2. `feat`: `reflection_strategy` injection seam
Adds `reflection_strategy: ReflectionLM | None` to `ReflectiveMutationProposer` and `gepa.optimize()`: any `ReflectionLM` implementation can own how reflective mutation calls the reflection model, taking precedence over the stateless default. Implementations providing only `reflect()` are called per task in the batch path (`getattr` fallback, same pattern as `adapter.batch_evaluate`); `reflect_many` remains the batching fast path.

This is the #329 Phase 2 entry point: session-based reflectors, coding-agent reflectors, and aggregating reflectors — **specifically ComBEE (#307), which maps onto this seam as a single `ComBEEReflectionLM` with internal map/reduce fan-out** — plug in without touching the proposer pipeline.

## 3. `docs`: parallel-proposals guide
New `docs/guides/parallel-proposals.md` covering the sampling/selection strategies (the surface replacing `num_parallel_proposals`), the optional `adapter.batch_evaluate` seam and its sequential fallback, budget-event granularity, and the `reflection_strategy` seam. Registered in mkdocs nav.

## Test plan
- [x] `tests/test_aime_prompt_optimization` replay passes on this branch (empirical behavior-identity check)
- [x] 358 tests pass locally (optional-dependency suites excluded, covered by CI)
- [x] ruff clean on all changed files (remaining findings pre-exist on the base branch)

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -75,6 +75,7 @@ jobs:
     name: Run Tests
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
     steps:
@@ -150,6 +151,7 @@ jobs:
     name: Build Package
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
     steps:
--- a/docs/docs/guides/parallel-proposals.md
+++ b/docs/docs/guides/parallel-proposals.md
@@ -1,0 +1,113 @@
+# Parallel Proposals: Sampling & Selection Strategies
+
+GEPA's reflective mutation loop can generate and evaluate **multiple candidate
+proposals per iteration**. This is controlled by two small, composable
+strategy objects rather than a numeric knob:
+
+- a **sampling strategy** decides *which (parent, minibatch) proposal tasks*
+  are attempted this iteration, and
+- a **selection strategy** decides *which of the resulting improvements* enter
+  the candidate pool.
+
+The defaults (`SingleMutationSampling` + `AllImprovements`) reproduce classic
+GEPA behavior exactly: one parent, one mutation per iteration. Existing runs
+do not change.
+
+!!! note "Replaces `num_parallel_proposals`"
+    Versions v0.1.2/v0.1.3 briefly shipped an experimental integer knob,
+    `EngineConfig.num_parallel_proposals`. It is replaced by the strategies on
+    this page (see [RFC #329](https://github.com/gepa-ai/gepa/issues/329)).
+
+## Sampling strategies
+
+```python
+from gepa.strategies.proposal_sampling import (
+    SingleMutationSampling,  # default: 1 parent, 1 mutation
+    SameParentSampling,      # n mutations of the same parent
+    IndependentSampling,     # n independent (parent, minibatch) tasks
+    PxNSampling,             # p parents x n mutations each
+)
+```
+
+| Strategy | Per iteration | Use when |
+|---|---|---|
+| `SingleMutationSampling()` | 1 parent, 1 mutation | Default; classic GEPA |
+| `SameParentSampling(n=4)` | 1 parent, `n` mutations | Exploit a strong parent with diverse edits |
+| `IndependentSampling(n=4)` | `n` parents, 1 mutation each | Explore the frontier broadly |
+| `PxNSampling(p=2, n=2)` | `p` parents × `n` mutations | Balanced explore/exploit |
+
+## Selection strategies
+
+```python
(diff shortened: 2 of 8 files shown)


Benzhang2004:
Reviewed the three follow-ups on top of the #369 stack — all solid, branch is green (510 passed, pyright clean, ruff unchanged). One substantive point on the reflection seam plus a naming suggestion; otherwise nits.

### `c51d347c` — restore child eval events + AIME pins ✅

Good catch on the regression — the batch `propose()` fired `on_evaluation_start/end` for the parent minibatch eval but dropped them for the child candidate eval. The restore has the right payload shape (`candidate_idx=None`, `parent_ids=[parent_idx]`) and matches the batched all-starts→eval→all-ends form the parent events already use.

The test rewrite is a real upgrade over the tautological `hasattr` checks — golden-prompt replay, budget-ledger invariants, explicit-vs-implicit default parity, and the child-event pin are exactly the right things to lock down. (Tiny nit: `test_budget_ledger_is_consistent` hardcodes `10` for the seed valset eval, so it's coupled to `valset[:10]`.)

### `d7f6c291` — `reflection_strategy` seam ⚠️

The wiring is clean (precedence + `getattr(..., "reflect_many")` fallback mirroring `batch_evaluate`), and this genuinely unlocks **custom reflection logic and stateless/aggregating reflectors like ComBEE**.

My one concern is that it oversells **stateful sessions**. `reflect()`/`reflect_many()` return `(proposal, next_lm)`, but the proposer discards `next_lm` and never reassigns `self._reflection_lm` (set once in `__init__`). So a fork-style session reflector (returns a new extended LM, leaves `self` immutable — the #329 contract) **never accumulates** — each iteration reflects from the original; a self-mutating one accumulates only at `N=1`, and with any multi-proposal sampler `reflect_many` runs N jobs against one shared instance. That's inherent to there being no pool yet (a batched `reflect_many` produces N `next_lm`s with no single one to carry). I'd either narrow the wording to "stateless/aggregating reflectors," or note that injected reflectors must be self-contained and stateful accumulation lands with the reflection-LM pool. The added tests use a `self`-returning stub, so they don't exercise this path.

Minor: `reflection_strategy` is on `gepa.optimize()` but not `optimize_anything()`, whereas sampling/selection are on both.

**Naming:** while we're here — the `ReflectionLM` protocol name collides with the `reflection_lm: LanguageModel` param (a `StatelessReflectionLM` *wraps* a `reflection_lm`, so the protocol isn't an LM). Since this is now surfacing in the public API via `reflection_strategy: ReflectionLM`, worth renaming before it sets. I'd suggest **`ReflectionStrategy`** (consistent with `SamplingStrategy`/`SelectionStrategy` and the `reflection_strategy` param → `StatelessReflectionStrategy`).

### `18ea20e8` — docs ✅

Clear and accurate on strategies / `batch_evaluate` / budget granularity. Two small fixes: "Both `gepa.optimize()` and `optimize_anything` accept the strategies directly" holds for sampling/selection but not `reflection_strategy`, and the ComBEE/session framing inherits the `next_lm` caveat.

PR #394, opened 2026-07-16
feat: ComBEE as a ReflectionLM — #307 re-hosted on the #369 seam (RFC #329 Phase 3)

LakshyAAAgrawal:
Re-hosts ComBEE ([arXiv:2604.04247](https://arxiv.org/abs/2604.04247)) from #307 onto the `ReflectionLM` protocol shipped in #369/v0.1.4. **ComBEE support was originally contributed by @nuglifeleoji in #307** — the algorithm, grouping math, augmented shuffle, and aggregation prompt are theirs (co-author credit on the commit); this PR moves it to the new seam, where it needs no engine or proposer changes.

## What it is

`ComBEEReflectionLM`: per component, duplicate the `n` reflection records `p` times and shuffle (seeded RNG), split into `k = ⌊√n⌋` groups, run one Level-1 (map) reflection call per group, then one Level-2 (reduce) call synthesizing the `k` intermediates into the final instruction. `n < 4` degenerates to standard single-call reflection.

```python
from gepa.proposer.reflective_mutation.combee import ComBEEReflectionLM

gepa.optimize(..., reflection_strategy=ComBEEReflectionLM("openai/gpt-5.1"), reflection_minibatch_size=40)
# or: ReflectionConfig(reflection_strategy=ComBEEReflectionLM(...)) in optimize_anything
```

This is the first non-trivial `ReflectionLM` implementation — the pattern session-based and coding-agent reflectors (#329 Phase 2) will follow.

## Behavior preservation (how the port was verified)

A 9-scenario characterization suite was written first and validated against **the original #307 implementation** (`_propose_new_texts_combee`), then run against this port: with identical RNG seed and scripted LM, old and new produce **byte-identical LM-call sequences and final texts** across all scenarios — grouping math (k=⌊√n⌋, base+remainder splits), augmented-shuffle duplication (each record exactly `p`× across map prompts), template routing (level-1 vs aggregation), degenerate fallback, multi-component runs, missing-component skips, custom templates, determinism/seed-sensitivity. The scenarios are pinned in `tests/test_combee_reflection_lm.py`, plus e2e tests driving ComBEE through both `gepa.optimize` and `optimize_anything` on a synthetic task (verifying the `k+1` cost contract and metadata flow end-to-end).

## Improvements over #307, per its review

- The `k=1` fallback and single-surviving-group paths **log** instead of falling back silently, and record prompts/raw outputs (the original left diagnostics empty on those paths).
- **Cost visibility**: `combee:<comp>:{k, num_lm_calls, mode, level1_prompts, level1_outputs}` and `combee:total_lm_calls` in `ReflectionProposal.metadata` → `on_proposal_end`, trackers, proposal metadata. The up-to-`(k+1)×` reflection spend is no longer invisible.
- `total_cost`/token delegation (str models via `gepa.lm.LM`; plain callables wrapped in `TrackingLM`) so **`max_reflection_cost` works** with ComBEE.
- `duplication_factor` and aggregation-template placeholders validated at construction.
- Conflicts with `adapter.propose_new_texts`/`custom_candidate_proposer` raise at construction (inherited from the #369 seam) instead of #307's silent `use_combee` no-op.

## Not carried over, deliberately

- The `use_combee`/`combee_duplication_factor`/`combee_aggregation_prompt`/`rng` parameters on `gepa.optimize` and `optimize_anything` — configuration lives on the strategy object; the front doors stay clean (seam-over-knob).
- The proposer-internal `_propose_new_texts_combee` method and its engine gating — obsoleted by the protocol.

## Test plan

- [x] 9-scenario characterization suite passes against BOTH the original #307 code and this port; call logs byte-identical
- [x] 17 new tests in `tests/test_combee_reflection_lm.py` (characterization + seam integration + 2 e2e)
- [x] 425 local tests pass; ruff clean; pyright 0 errors on the new module
- [x] Docs: `guides/combee.md` rewritten for the `reflection_strategy` interface, registered in nav

Closes #307.


LakshyAAAgrawal:
**Paper-fidelity audit** (against [arXiv:2604.04247](https://arxiv.org/abs/2604.04247) directly, not just #307):

**Faithful:**
- §3.1 hierarchy: the paper's default instantiation is exactly the two-level scan implemented here — `k = ⌊√n⌋` groups so "the first level generates context updates based on n/k = √n reflections per group, while the second level aggregates k = √n updates". Deeper trees are permitted by the paper only under non-default k.
- §3.2 augmented shuffle: duplicate each reflection `p` times (paper default p=2) **then** shuffle — matches (`records * p` → seeded shuffle → contiguous split).
- Prompt contents, uneven-division handling (base+remainder), and the small-n fallback are **unspecified by the paper**; #307's choices are inherited unchanged and marked as implementation decisions, not paper requirements.

**Not implemented (in #307 nor here): §3.3, the dynamic batch size controller** — the paper's third mechanism (profile `T_epoch(bs) = A·bs^(−α)`, select the plateau batch size at marginal threshold τ=1.6%). In GEPA terms this is *adaptive `reflection_minibatch_size`*, which belongs at the batch-sampler seam rather than inside a `ReflectionLM` — filed as a follow-up issue rather than scope-crept into this PR.

**One deliberate divergence from #307's wiring, disclosed:** #307 passed the engine's seed-derived RNG (shared with the candidate selector) into ComBEE's shuffles, so shuffles perturbed the selection stream and vice versa. The strategy object uses an independent `Random(0)` by default — decoupled and reproducible, but it means full optimization runs are **not** bit-reproducible against #307, and varying `gepa.optimize(seed=...)` does not vary the shuffle stream (documented in the guide: pass `rng=random.Random(your_seed)` for seed-varied experiments). The per-proposal algorithm remains byte-identical under an injected RNG, per the characterization harness.

LakshyAAAgrawal:
@nuglifeleoji, would you be able to review this?

Once the PR is ready, we can merge this PR into #307, and then merge that branch into main so all of your original commits are preserved as well.

Benzhang2004 (commented):
lgtm. merging.

PR #400, opened 2026-07-22
fix: support batch eval and old config

Shangyint:
adds evaluate batch support and all legacy arguments: now the new api should be a strict superset of the old API!!

### DIFF
--- a/src/gepa/api.py
+++ b/src/gepa/api.py
@@ -157,7 +157,7 @@ def optimize(
     - reflection_minibatch_size: The number of examples to use for reflection in each proposal step. Defaults to 3. Only valid when batch_sampler='epoch_shuffled' (default), and is ignored otherwise.
     - perfect_score: The perfect score to achieve.
     - reflection_prompt_template: The prompt template to use for reflection. Can be either a string (applied to all components) or a dict mapping component names to their specific templates. If not provided, GEPA will use the default prompt template (see [InstructionProposalSignature](src/gepa/strategies/instruction_proposal.py)). Each prompt template must contain the following placeholders, which will be replaced with actual values: `<curr_param>` (will be replaced by the instructions/component to evolve) and `<side_info>` (replaced with the inputs, outputs, and feedback generated with current instruction). When using a dict, components without a specified template will use the default template. This will be ignored if the adapter provides its own `propose_new_texts` method.
-    - custom_candidate_proposer: Optional custom function for proposing new candidates. If provided, this will be used instead of the default LLM-based reflection approach. Cannot be used if adapter provides `propose_new_texts`. Signature: `(candidate, reflective_dataset, components_to_update) -> dict[str, str]`.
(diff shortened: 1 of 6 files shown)


LakshyAAAgrawal:
Thanks — this cleanly restores `batch_evaluator` to the new front door and the `GEPAConfig` bridge (`_from_legacy_config`) round-trips correctly end-to-end. A few things to flag before this closes out the #346 regressions:

**Still open from #346 (not addressed here):**
- **Wrong "best" on budget termination.** On `BudgetExhausted` with `run_dir=None` (the default), `GepaEngine.run` falls back to `Result(best_candidate=server.best_candidate, ...)`, which is the argmax over individual `(candidate, example)` evals — not the val-aggregate/Pareto best. Multi-task/generalization runs can return the wrong candidate. Suggest reloading the Pareto best even when `run_dir` is unset, or making the server cap soft (between-iteration) like the launcher stopper.
- **`custom_candidate_proposer` signature break, silently swallowed.** `reflective_mutation.py:157-162` now calls the proposer with `metadata=`, but `api.py:160` still documents the 3-arg `(candidate, reflective_dataset, components_to_update)` form. Existing proposers raise `TypeError`, which `_propose_texts_batch_safe` (`reflective_mutation.py:234-249`) catches, retries, and drops to `None` → every proposal silently discarded, budget spent, run reports success. This is on the shared core path (affects `gepa_launcher` too). Suggest accepting 3-arg proposers (or updating the doc) and making the failure loud.
- **Failed-eval budget undercount.** Evaluator exceptions bypass `budget.record()`, so failures do not count toward the cap; the new `evaluate_batch` has the same shape (records only after the user call returns).
- **`opt_state` warm-start** is now forwarded on the batch path but still dropped on the per-pair path (the `evaluator` closure in `gepa.py` swallows `**kwargs`).
- Legacy imports (`from gepa.optimize_anything import GEPAConfig / EngineConfig / make_litellm_lm`) still raise `ImportError`, and the return type is `Result`, not `GEPAResult`.

**Introduced by this PR (batch path only):**
- **All-or-nothing batch failure.** In `evaluate_examples`, if the batch fn raises anything other than `BudgetExhausted`, every example in the stage is marked `0.0`. The per-pair path isolates a single failure to a single `0.0`. A transient batch error can zero a whole valset, and those `0.0`s enter Pareto state as real scores — risking wrongly rejecting a good candidate or corrupting the frontier. No `_gepa_transient_failure` handling on this path.
- **Softer budget cap.** `evaluate_batch` checks budget once up front, then records all `len(pairs)`, so a stage can overshoot `max_evals` by up to `stage_size − 1` (e.g. ~149 for a 150-example valset). The per-pair path stopped precisely. Worth documenting explicitly if intended.
- Minor: `_normalize_result` forces `dict(info)`, so an evaluator returning `(score, non_dict_info)` now raises where it previously passed through.

Shangyint:
Thanks. Going through them one by one - some are features rather than bugs ;)

Shangyint:
Fixed: 
Wrong "best" on budget termination. 
custom_candidate_proposer signature break, silently swallowed. 
Failed-eval budget undercount.
opt_state warm-start 
All-or-nothing batch failure. 

Non-problem: 
Legacy imports: gepa specific configs should stay in gepa_launcher
Softer budget cap: this is current behavior, documented, and makes sense.
dict(info) strictness: this is enforced, in other places too!

PR #410, opened 2026-07-28
fix(oa): avoid AutoResearch subprocess deadlocks

sh-patterson:
## Summary

- prevent AutoResearch agent subprocesses from deadlocking when stdout or stderr exceeds the OS pipe buffer
- terminate the whole POSIX subprocess group when the no-eval or exhausted-budget watchdog fires
- preserve captured output and existing watchdog reason markers
- retain direct-process termination as the Windows fallback

## Root cause

`AutoResearchEngine._run_claude` launched the agent with `stdout=PIPE` and
`stderr=PIPE`, but did not drain either stream until `poll()` reported that the
process had exited. A sufficiently verbose agent could therefore fill a pipe,
block before exiting, and leave the parent polling forever.

The watchdog paths also terminated only the immediate `claude` process. Child
processes could survive the watchdog and continue running after the engine had
returned.

## Implementation

Agent output is redirected into temporary files, removing pipe backpressure
without adding reader threads. On POSIX, the agent starts in a new session and
watchdog termination signals its process group, escalating from `SIGTERM` to
`SIGKILL` after the existing grace period. Windows keeps direct
`terminate()`/`kill()` behavior.

## Validation

- `python -m pytest -q tests/test_optimize_anything_autoresearch_engine.py`
  - 11 passed
- `python -m pytest -q tests/test_optimize_anything_*.py`
  - 39 passed
- `python -m ruff check src/gepa/oa/engines/autoresearch.py tests/test_optimize_anything_autoresearch_engine.py`
  - passed
- `python -m pyright src/gepa/oa/engines/autoresearch.py`
  - 0 errors

The new regression coverage exercises:

- simultaneous 256 KiB stdout and stderr output
- POSIX descendant-process termination
- preserved no-eval and budget-exhaustion diagnostics
- the Windows direct-process fallback

### DIFF
--- a/src/gepa/oa/engines/autoresearch.py
+++ b/src/gepa/oa/engines/autoresearch.py
@@ -20,6 +20,7 @@ from __future__ import annotations
 import json
 import os
 import shutil
+import signal
 import subprocess
 import tempfile
 import time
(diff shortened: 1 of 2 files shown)


Shangyint:
Thanks! @sh-patterson. This looks good to me. Happy to merge it after you mark this as ready.

Issue #409, opened 2026-07-24
Code review findings — main @ f919db0a (2026-07-24)


LakshyAAAgrawal:
## Code review findings — `main` @ `f919db0a622e2e9f9204779b81fe00cc1b2d808f`

**Date:** 2026-07-24
**Reviewed:** full diff `cdfcc8fb703eb3272b817b1d6d7d07561d6830cc..f919db0a` (releases 0.1.3 → 0.1.4+: the `optimize_anything` rewrite + new `oa/` engine-pluggable module, engine/state refactor, ReflectionLM + parallel proposals, sandbox-on-by-default, new `gepa_launcher.py`), then re-scoped to the codebase as it currently stands at that HEAD.

**Baseline health:** all imports resolve, the legacy `optimize_anything` shim is complete, and the full test suite passes (608 passed, 4 skipped). No crash/P0 bugs. The default path (single proposal, `"gepa"` engine, no agent-state, strict acceptance) is behaviorally intact; cache/determinism/resume are sound; no test-set leakage and no command-injection/path-traversal in the agent engines.

The findings below are ordered most → least important.

---

### 1. `optimize_anything()` returns the wrong result type for no-config callers  (P1, silent, normal usage)
`src/gepa/optimize_anything.py:169-221` — the legacy `GEPAResult` is only returned when a `GEPAConfig` instance is passed. Calling it the documented simple way (`optimize_anything(seed, evaluator=fn)` with no config, or `config=None`) now returns the new `Result`. `.best_candidate` still works, but `.val_aggregate_scores`, `.best_idx`, `.candidates`, `.to_dict()` all raise `AttributeError`. Most likely to break existing user code silently.

### 2. Eval-budget double-charge in the batch-eval fallback  (P1, silent, corrupts runs)
`src/gepa/oa/eval_server.py:371-390` (+ failure-record loop `:294-300`) — when a `batch_evaluate` fn raises, the server records N failure ticks and re-raises; `evaluate_examples` then catches and retries per-pair, charging N again. Verified: 3 examples with a raising batch fn → real evaluator invoked 3×, but `budget.used == 6`. Any stage where the batch fn intermittently fails silently burns ~2× budget and terminates the run early. The code comment calls this intentional, but charging N for a call that produced zero evals *plus* N for the retry is an accounting defect.

### 3. `evaluate_batch` raises mid-record instead of honoring its documented "overshoot"  (P2, discards work)
[...]

PR #411, opened 2026-07-28
fix(oa): return GEPAResult when optimize_anything config is omitted

rishiskhare:
## What & why

`optimize_anything(seed, evaluator=fn)` with no config was returning the new `Result`, so `.val_aggregate_scores`, `.best_idx`, `.candidates`, and `.to_dict()` broke for v0.1.x callers (#409 finding 1). Only an explicit `GEPAConfig` took the `GEPAResult` path.

Omitted config now results in `GEPAResult` the same way, and `test_set` is rejected on that path. Explicit `OptimizeAnythingConfig` still returns `Result`.

Part of #409

## Tests

- `test_no_config_returns_gepa_result`
- `test_no_config_rejects_test_set`
- `uv run pytest tests/test_optimize_anything_gepa_config.py`

### DIFF
--- a/src/gepa/oa/ensemble.py
+++ b/src/gepa/oa/ensemble.py
@@ -80,15 +80,20 @@ def optimize_anything_from_task(
         config = OptimizeAnythingConfig()
     if config.name is None:
         config = replace(config, name=task.name)
-    return optimize_anything(
-        task.seed_candidate,
-        evaluator=evaluate,
-        dataset=task.train_set,
-        valset=task.val_set,
-        objective=task.objective,
-        background=task.background,
-        test_set=task.test_set,
-        config=config,
+    # config is always a concrete OptimizeAnythingConfig here, so
+    # optimize_anything returns Result (not the GEPAResult union arm).
+    return cast(
+        Result,
+        optimize_anything(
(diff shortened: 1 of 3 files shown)


LakshyAAAgrawal:
Thanks a lot for working on this @rishiskhare. I made some changes to your PR and merged #420 with all credits due to you (added you as a co-author).

rishiskhare:
Thanks @LakshyAAAgrawal! I would encourage you to continue to build oa-omni as the primary interface for self-improvement in production. The community is looking for plug-and-play solutions for recursive self-improvement, and this PyPi library has been the clear winner in terms of adoption across enterprises as a production-ready package. The other RSI strategies (meta-harness, auto-research) have repositories that don't have the stability and practicality of this API. I would strongly encourage you to make this gepa optimize-anything library the primary surface for other complementary RSI approaches for this reason - it would be awesome if you could broaden the scope of the meta-harness implementation to adopt Codex (#417), as well as explore integrate first-class approaches to make other optimizers easy to use from this library itself. I think with this strategy, this library could turn into the defacto solution for self-improving AI systems at large, not just scoped for the GEPA method itself.

PR #420, opened 2026-08-03
refactor(oa): unify optimize_anything on a single GEPAResult return

### LINKED ISSUE #409: Code review findings — main @ f919db0a (2026-07-24)

LakshyAAAgrawal:
## Credit

This change is directly based on work by [Rishi Khare](https://github.com/rishiskhare) (@rishiskhare).

He filed the original diagnosis as finding 1 of #409: `optimize_anything(seed, evaluator=fn)` with no config was returning the new `Result`, so `.val_aggregate_scores`, `.best_idx`, `.candidates`, and `.to_dict()` broke for v0.1.x callers. He then proposed the first fix in #411: omitted config should return `GEPAResult`.

This branch starts from his #411 commits (his original commit is the first commit here), then unifies the public return to a single `GEPAResult` so the same pool surface is available for every config path, including explicit `OptimizeAnythingConfig`. He is a co-author on the follow-up commits.

## Summary

`optimize_anything` returned `Result | GEPAResult` depending on the config type. That split the public surface and forced a union/cast at every internal boundary:

- Callers with an explicit `OptimizeAnythingConfig` got the lean `Result` and could not reach `candidates` / `best_idx` / `val_aggregate_scores` / `to_dict()` (#409).
- Only the gepa/legacy path returned the rich `GEPAResult`.
- `test_set` was rejected on the `GEPAResult` path because that type had nowhere to store held-out scores.
- The union required a `cast(Result, ...)` in the ensemble helper (from #411).

This collapses everything to **one return type**: `optimize_anything` always returns a `GEPAResult`.

The universal core is always populated: `best_candidate`, `best_score`, `total_evals`, `eval_log`, `metadata` (the old lean accessors) **plus** the `candidates` / `best_idx` / `val_aggregate_scores` / `to_dict()` pool surface. The gepa engine fills the full candidate pool and Pareto data; other engines return a single-candidate snapshot with the gepa-only fields left `None`/empty. Held-out `test_set` scores now land in `metadata` on **any** engine.

## Changes

- **`GEPAResult`** gains `eval_log` + `metadata` fields and `best_score` + `total_evals` properties. Additive and backward compatible: `to_dict()` / `from_dict()` are **untouched**, so the frozen JSON serialization contract stays byte-stable for downstream consumers (`.candidates`/`.best_idx`/`.val_aggregate_scores`/`.to_dict()`).
- `total_evals` prefers the eval-server count (`budget.used`) over GEPA core's `total_metric_calls`.
- The mutable engine-internal `Result` is **unchanged**. `optimize_anything`'s core is extracted into `_optimize_to_result` (returns the mutable `Result`) and projected onto `GEPAResult` by `_to_gepa_result` at the public boundary.
- Ensemble helpers call `_optimize_to_result` directly, so they keep threading the mutable `Result` across stages. That removes the need for the `cast(Result, ...)` in `optimize_anything_from_task`.

## Compatibility

- The T1 `GEPAResult` public shape is only extended, never changed.
- Explicit-config callers keep every accessor they had and additionally gain the pool surface.
- Legacy `GEPAConfig` in / `GEPAResult` out is preserved.
- `metadata` no longer nests `gepa_result`; the return value *is* the pool (`result.candidates`).

## Testing

- Coverage for the unified contract: real no-config path (#409 / #411), explicit `OptimizeAnythingConfig`, single-candidate generic projection, and `test_set` on every path.
- `to_dict()` still omits runtime OA fields (`eval_log`, `metadata`, `eval_server_calls`).

Builds on #411; supersedes its union/cast end state; fixes #409.

### DIFF
--- a/.claude/skills/gepa-optimize-anything/SKILL.md
+++ b/.claude/skills/gepa-optimize-anything/SKILL.md
@@ -140,7 +140,7 @@ a "best candidate" that looks fine but is barely optimized.
   themselves how to spend eval calls, so treat it as a floor.)
 
 After the run, check how many proposals actually happened (on the gepa backend,
-`result.metadata["gepa_result"]` holds every candidate; with `engine.write_agent_state=True` the
+`result.candidates` is the full pool; with `engine.write_agent_state=True` the
 `run_dir/iterations/` tree shows each one). **If it stopped after one proposal, the budget was too
 low** — raise it and rerun.
 
@@ -248,7 +248,7 @@ These silently degrade *results* — skim before launching:
 
 ## Reference files (load as needed)
 - `references/api.md` — `OptimizeAnythingConfig`, the backends and their typed `engine_config`
-  options, the three modes, the LM protocol, budget/cost semantics, `Result` shape, and the
+  options, the three modes, the LM protocol, budget/cost semantics, `GEPAResult` shape, and the
   composition/pipeline helpers.
 - `references/writing_evaluators.md` — the `(score, info)` contract, `oa.log()`/`capture_stdio`,
   LLM-as-judge scoring, multi-objective via `info["scores"]`, N>1 averaging, feedback design.
--- a/.claude/skills/gepa-optimize-anything/references/api.md
+++ b/.claude/skills/gepa-optimize-anything/references/api.md
@@ -4,7 +4,7 @@ The public entry point is `gepa.optimize_anything.optimize_anything`.
 
 ## Signature
 ```python
-from gepa.optimize_anything import optimize_anything, OptimizeAnythingConfig
+from gepa.optimize_anything import GEPAResult, optimize_anything, OptimizeAnythingConfig
 
 result = optimize_anything(
     seed_candidate: str | None = None,   # starting text; None = seedless (engine bootstraps from objective/background)
@@ -17,7 +17,7 @@ result = optimize_anything(
     background: str | None = None,       # long-form rules/constraints/domain notes
     test_set: list | None = None,        # reporting-only: seed + final candidate scored here at the end
     config: OptimizeAnythingConfig | None = None,
-) -> Result
+) -> GEPAResult
 ```
 Examples in `dataset`/`valset`/`test_set` are opaque — any object your `evaluator` understands.
 `seed_candidate` is a **single string** at this API (multi-component `dict[str, str]` candidates
@@ -220,21 +220,40 @@ take caller-owned `EvalServer`s — one per config (one shared server for
 eval through their own server. Related: the autoresearch backend's `handoffs` key materializes
 prior-stage artifacts into the agent's work dir for hand-rolled sequential compositions.
 
-## `Result`
+## `GEPAResult`
+
+`optimize_anything` always returns a `GEPAResult` (the same type as the legacy
+`gepa.gepa_launcher` path). The mutable engine-internal `Result` is still what
+engines and ensemble helpers thread; it is not the public return.
+
 ```python
-result.best_candidate   # str
+result.best_candidate   # str (or dict[str, str] for multi-component seeds)
 result.best_score       # float (on valset/selection set)
-result.total_evals      # int
+result.total_evals      # int — eval-server call count (budget.used)
 result.eval_log         # list[dict]
 result.metadata         # dict (verified keys):
-#   "gepa_result"            full GEPAResult (all candidates + per-instance val scores) — gepa engine only
 #   "test_score"             avg over test_set         } only present
 #   "test_scores"            per-example dict          } if you passed
 #   "baseline_test_score"    seed avg over test_set    } a test_set
 #   "baseline_test_scores"   seed per-example dict     }
 #   "budget", "total_cost", "adapter_cost", "wall_time", "engine", "output_dir", "progress_log"
+result.candidates            # full candidate pool (length 1 on non-gepa engines)
+result.best_idx              # index of the highest val score
+result.val_aggregate_scores  # per-candidate average validation score
+result.to_dict()             # JSON-safe pool snapshot (does not include eval_log / metadata)
 ```
+
+The candidate pool lives on the result itself. `metadata` does **not** contain a nested
+`gepa_result`. `total_evals` is the eval-server count; `total_metric_calls` is GEPA core's
+own counter and may differ (e.g. after `BudgetExhausted`). `to_dict()` / `from_dict()` omit
+`eval_log`, `metadata`, and `eval_server_calls` so the frozen JSON pool contract stays
+byte-stable.
+
 The gepa engine also writes `run_dir/` artifacts, and the eval server writes
-`output_dir/summary.json`. The `gepa_result` in metadata is the richest artifact — keep it for
-post-hoc analysis. (If the best candidate equals the seed, the test scores are reported from the
+`output_dir/summary.json`. (If the best candidate equals the seed, the test scores are reported from the
 seed's single scoring pass rather than re-scored.)
+
+**Migration:** callers that used `isinstance(result, Result)` or
+`result.metadata["gepa_result"]` after `optimize_anything(...)` should switch to
+`GEPAResult` and `result.candidates`. Ensemble helpers (`optimize_sequential`, …)
+still return the mutable `Result`.
--- a/.gitignore
+++ b/.gitignore
@@ -217,3 +217,7 @@ runs/
 outputs/
 gepa_terminus/
 archive/
+
+# Local MLflow tracking artifacts (TrackingConfig(use_mlflow=True))
+mlflow.db
+mlruns/
--- a/src/gepa/core/result.py
+++ b/src/gepa/core/result.py
@@ -1,7 +1,7 @@
 # Copyright (c) 2025 Lakshya A Agrawal and the GEPA contributors
 # https://github.com/gepa-ai/gepa
 
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any, ClassVar, Generic
 
 from gepa.core.adapter import RolloutOutput
@@ -61,6 +61,15 @@ class GEPAResult(Generic[RolloutOutput, DataId]):
     # This is the internal dict key used to wrap str seed_candidates.
     _str_candidate_key: str | None = None
 
+    # Universal-core run data, populated on every run regardless of engine.
+    # Generic (non-gepa) engines return a single-candidate snapshot but still
+    # carry their eval log and free-form metadata (wall time, budget, cost,
+    # held-out test scores, ...). These are runtime conveniences and are
+    # intentionally NOT part of ``to_dict()`` — that stays a JSON-safe pool
+    # snapshot for the frozen serialization contract.
+    eval_log: list[dict[str, Any]] | None = None
+    metadata: dict[str, Any] = field(default_factory=dict)
+
     _VALIDATION_SCHEMA_VERSION: ClassVar[int] = 2
 
     # -------- Convenience properties --------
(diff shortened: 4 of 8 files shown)


LakshyAAAgrawal:
Rebased onto current `main` (keeps #423's legacy `max_workers` / `parallel=False` mapping) and addressed the review:

- Dropped the accidental `mlflow.db`; gitignored `mlflow.db` / `mlruns/`
- `total_evals` now prefers `eval_server_calls` (`budget.used`) over `total_metric_calls`; `to_dict()` still omits the runtime fields
- `eval_log` defaults to `[]` instead of `None`
- Real `config=None` coverage (the previous test passed an explicit `OptimizeAnythingConfig`)
- Skill + `references/api.md` now document `GEPAResult` / `result.candidates` instead of `metadata["gepa_result"]`

CI should be green on the rebased HEAD (`c6288822`).

PR #425, opened 2026-08-12
fix(cache): namespace valset evaluations so they cannot reuse trainset rollouts

### LINKED ISSUE #424: Evaluation cache conflates trainset and valset ids: valset scores are silently served from trainset rollouts
## Summary

When `valset` is passed as a separate dataset (the documented **Generalization** mode), GEPA can record a valset score for an example it never evaluated, reusing the rollout of a **different** example from the trainset.

`ListDataLoader` ids are list positions, the trainset and the valset get independent loaders each numbering from `0`, and `EvaluationCache` is keyed `(candidate_hash, example_id)` — so trainset item *i* and valset item *i* collide on one cache key.

This does not affect the modes where `valset=None`: there `val_loader is train_loader`, so the shared id space is correct and the cache does exactly what it should. It only bites in the mode whose purpose is measuring generalization to held-out data.

## Reproduction

```python
"""GEPA: valset scores are served from trainset rollouts when a separate valset is passed.

Note the mock examples deliberately carry no "id" key: GEPA's example ids here are purely
list positions. The "train-N"/"val-N" strings are the example *contents*, named after their
position only so the collision is legible in the output.
"""
import collections
import json

from gepa.core.state import EvaluationCache

TRAIN = [{"content": f"train-{n}"} for n in range(12)]
VAL = [{"content": f"val-{n}"} for n in range(6)]
SLOT_OF = {example["content"]: slot for slot, example in enumerate(VAL)}

# candidate -> valset slot -> content of the example that actually produced that slot's score
produced_by = collections.defaultdict(dict)


def tag_of(candidate):
    raw = next(iter(candidate.values()))
    try:
        return json.loads(raw)["text"]
    except Exception:
        return str(raw)


def ran_content(cached_output):
    """The evaluator records the example it ran; find that marker in the cached output."""
    parts = cached_output if isinstance(cached_output, tuple) else (cached_output,)
    return next((part["ran"] for part in parts if isinstance(part, dict) and "ran" in part), "?")


_get_batch = EvaluationCache.get_batch


def get_batch(self, candidate, example_ids):
    """Valset lookups come through here; a hit fills a slot without running anything."""
    cached, uncached = _get_batch(self, candidate, example_ids)
    for slot, entry in cached.items():
        produced_by[tag_of(candidate)][slot] = ran_content(entry.output)
    return cached, uncached


EvaluationCache.get_batch = get_batch

from gepa.optimize_anything import EngineConfig, GEPAConfig, ReflectionConfig, optimize_anything


def evaluator(candidate, example):
    tag = json.loads(candidate)["text"]
    content = example["content"]
    if content in SLOT_OF:
        produced_by[tag][SLOT_OF[content]] = content
    # monotonically improving candidates, so children are accepted and trigger valset evaluations
    score = 0.1 * (int(tag.split("-")[-1]) + 1 if tag != "seed" else 0)
    return score, {"ran": content}


counter = iter(range(999))
optimize_anything(
    seed_candidate=json.dumps({"text": "seed"}),
    evaluator=evaluator,
    dataset=TRAIN,
    valset=VAL,
    objective="repro",
    config=GEPAConfig(
        engine=EngineConfig(
            max_metric_calls=200,
            max_candidate_proposals=5,
            display_progress_bar=False,
            cache_evaluation=True,
        ),
        reflection=ReflectionConfig(
            reflection_lm=lambda _prompt: json.dumps({"text": f"cand-{next(counter)}"}),
            reflection_minibatch_size=4,
        ),
    ),
)

print("\nContent of the example that actually produced each valset slot's score")
print("(* = a trainset example's rollout, reused for a valset slot)\n")
print("  " + "candidate".ljust(10) + "".join(f"slot {slot}".ljust(11) for slot in range(len(VAL))))
for tag, slots in produced_by.items():
    cells = []
    for slot in range(len(VAL)):
        content = slots.get(slot, "-")
        cells.append((content + ("*" if content.startswith("train-") else "")).ljust(11))
    reused = sum(1 for value in slots.values() if value.startswith("train-"))
    print("  " + tag.ljust(10) + "".join(cells) + (f"  {reused}/{len(VAL)} reused" if reused else ""))
```

Output:

```
Content of the example that actually produced each valset slot's score
(* = a trainset example's rollout, reused for a valset slot)

  candidate slot 0     slot 1     slot 2     slot 3     slot 4     slot 5     
  seed      val-0      val-1      val-2      val-3      val-4      val-5      
  cand-0    val-0      train-1*   val-2      val-3      val-4      train-5*     2/6 reused
  cand-1    val-0      val-1      train-2*   train-3*   val-4      val-5        2/6 reused
  cand-2    train-0*   val-1      val-2      val-3      train-4*   val-5        2/6 reused
  cand-3    val-0      val-1      val-2      train-3*   val-4      val-5        1/6 reused
  cand-4    train-0*   val-1      val-2      val-3      val-4      train-5*     2/6 reused
```

The seed is clean: it is evaluated on the valset before any minibatch rollouts exist to collide with. Every candidate after it has valset slots filled by trainset rollouts, and the collision is plainly positional — valset slot 1 is filled by the rollout of trainset position 1, slot 5 by trainset position 5, and so on. Each of those candidates still ends up with a full six-value valset score vector, and nothing in the run output indicates that two of the six were never evaluated.

## Root cause

1. `ListDataLoader.all_ids()` returns `range(len(items))`, so ids are list positions — `src/gepa/core/data_loader.py:56-57`.
2. Two independent loaders are built when a valset is supplied — `src/gepa/gepa_launcher.py:1381-1382`:
   ```python
   train_loader = ensure_loader(effective_dataset)
   val_loader = ensure_loader(valset) if valset is not None else train_loader
   ```
3. A single `EvaluationCache` serves both, keyed `(candidate_hash, example_id)` — `src/gepa/core/state.py:30,72-87`.
4. The valset path reads the cache (`src/gepa/core/engine.py:306`) while the minibatch path only writes it (`src/gepa/proposer/reflective_mutation/reflective_mutation.py:386,650`). The contamination is therefore one-directional: valset results are replaced by trainset results, never the reverse.

Expected collisions per candidate are about `minibatch_size × |valset| / |trainset|`, so this grows with minibatch size and with the valset fraction.

## Why it matters

The fabricated per-instance scores do not just distort the reported aggregate — they feed candidate selection. `prog_candidate_val_subscores`, `pareto_front_valset` and `program_at_pareto_front_valset` all store the cached value, and `select_program_candidate_from_pareto_front` (`src/gepa/gepa_utils.py:107`) samples parents proportionally to the number of valset instances a candidate holds on the frontier. Each contaminated instance is an extra selection ticket.

The bias is systematically upward, for two compounding reasons: the reused value is the candidate's *own* minibatch result, i.e. the data the reflection step just optimized against, and a child only reaches the valset stage if those minibatch scores were good. So it is in-sample signal, filtered for high scores, injected into the held-out measurement.

In one of our runs (68 train / 18 val, minibatch 10), 11 of 108 valset slot evaluations were substituted this way. Their mean score was 0.895 against 0.659 for the genuinely-run valset evaluations; one candidate's reported valset average was inflated by +0.105 (0.592 → 0.697); and 10 of the 11 placed their candidate on the Pareto front for an instance it never ran, including one instance where the candidate was the *sole* frontier occupant purely on the strength of a different example's rollout.

## Suggested fix

Disambiguate the cache key when the two loaders are not the same object, e.g. namespace by split or by loader identity:

```python
self._cache[(_candidate_hash(candidate), split_key, example_id)] = ...
```

The newer `oa` eval-server path already does the equivalent, defaulting ids to `f"{split_name}_{i}"` (`src/gepa/oa/eval_server.py:205-215`); the core loader/cache layer just hasn't caught up. Alternatives: give `ListDataLoader` an id prefix/offset, or have `ensure_loader` assign each loader it creates a distinct namespace.

An assertion that `val_loader is train_loader` or that their id sets are disjoint would have caught this at startup.

## Workarounds

- `cache_evaluation=False` avoids it completely, at the cost of re-running evaluations — expensive when a single evaluation takes minutes.
- On the released `0.1.4` API, passing a custom `DataLoader` whose `all_ids()` returns stable unique ids (we use dataset-level case ids) fixes it: `ensure_loader` duck-types it through and the valset state is all `dict[DataId, ...]`. On `main` this no longer works through `optimize_anything`, because `EvalServer.__init__` iterates the dataset directly (`src/gepa/oa/eval_server.py:211`) and a loader is not iterable.

## Environment

- Reproduced on `main` (`pip install git+https://github.com/gepa-ai/gepa@main`) and on released `0.1.4`.
- Python 3.10.

blerchin:
Fixes #424.

## The bug

Data loaders identify examples by position (`ListDataLoader.all_ids()` returns `range(len(items))`), and a separate valset gets its own loader that also numbers from zero. `EvaluationCache` is keyed `(candidate_hash, example_id)`, the valset path reads it (`core/engine.py`) and the minibatch path only writes it (`reflective_mutation.py`), so a valset lookup could be served by the rollout of the unrelated trainset example at the same position.

The result is a full valset score vector for candidates that never ran some of those examples — and because those per-instance scores land in `prog_candidate_val_subscores` and `program_at_pareto_front_valset`, they also feed parent selection, which samples proportionally to frontier appearances.

This only affects the **Generalization** mode (`dataset=<list>, valset=<list>`). With `valset=None` the launcher already passes the same loader for both, so the shared id space is correct and the cache is doing the right thing.

## The fix

Cache reads and writes take a `split`. The launcher decides it once, where both loaders are in scope:

```python
valset_cache_split = TRAINSET_CACHE_SPLIT if val_loader is train_loader else VALSET_CACHE_SPLIT
```

- **Distinct valset** → valset entries get their own namespace and can no longer collide with minibatch entries.
- **`valset=None`** → both sides use the same split, the ids genuinely refer to the same examples, and cached rollouts are still shared across minibatch and valset evaluation. This is deliberate: namespacing unconditionally would have cost real evaluations in the default mode.

`GEPAEngine` also derives the split itself when it is not passed, so callers that build the engine directly get the safe behaviour without changes.

### Backwards compatibility

`_key()` returns the previous two-element key for the default split, so evaluation caches persisted by earlier versions stay readable on resume. Valset entries written before this change are simply not reused (they are the contaminated ones).

The only signature changes are new trailing parameters with defaults, on `EvaluationCache.get/put/get_batch/put_batch/evaluate_with_cache_full` and `GEPAState.cached_evaluate/cached_evaluate_full`.

## Tests

`tests/test_valset_cache_split.py`:

- `test_valset_slots_are_never_filled_by_trainset_rollouts` — the regression test. Runs a short optimization with a distinct valset and asserts every valset slot's score came from the valset example that owns it. Against unpatched `main` it fails with the leak spelled out:

  ```
  AssertionError: valset slots served by trainset rollouts:
    {'cand-0': {1: 'train-1', 5: 'train-5'}, 'cand-1': {2: 'train-2', 3: 'train-3'},
     'cand-2': {0: 'train-0', 4: 'train-4'}, 'cand-3': {3: 'train-3'},
     'cand-4': {0: 'train-0', 5: 'train-5'}}
  ```

  The collision is positional, and GEPA's own log reported `coverage 6 / 6` for those candidates — which is what made this silent.

- `test_shared_valset_still_reuses_trainset_rollouts` — guards the other direction: with `valset=None`, cache hits must still occur. This one **passes on unpatched `main` too**, so it fails if the fix is ever tightened into "namespace everything".

- Two unit tests covering the key isolation and the preserved legacy key shape.

The mock examples deliberately carry no `id` key and use `{"content": "train-N"}`, so the output distinguishes GEPA's example ids (positions) from example contents.

`uv run pytest`: 673 passed, 4 skipped. `ruff check` and `pyright` clean on the touched files. One existing test double in `tests/proposer/test_merge.py` needed the new optional parameter.

### DIFF
--- a/.claude/skills/gepa-optimize-anything/SKILL.md
+++ b/.claude/skills/gepa-optimize-anything/SKILL.md
@@ -157,7 +157,8 @@ low** — raise it and rerun.
 - If you **opt in** to evaluation caching (`engine_config={"engine": {"cache_evaluation": True}}` on
   the gepa backend — it is **off by default**), be aware `max_evals` then counts only cache *misses*:
   a converged search can keep proposing cache-hitting candidates without consuming eval budget, so
-  `stop_at_score`/`max_token_cost` become mandatory, not optional.
+  `stop_at_score`/`max_token_cost` become mandatory, not optional. A distinct `valset` is cached
+  separately from the trainset; `valset=None` still reuses minibatch rollouts.
 
 ## Minimal working example
 The example optimizes a system prompt for concreteness, but the **shape is identical** for any
--- a/.claude/skills/gepa-optimize-anything/references/api.md
+++ b/.claude/skills/gepa-optimize-anything/references/api.md
@@ -106,7 +106,9 @@ engine_config={
         "frontier_type": "hybrid",       # "instance" | "objective" | "hybrid" (default) | "cartesian"
         "candidate_selection_strategy": "pareto",      # | "current_best" | "epsilon_greedy" | "top_k_pareto"
         "acceptance_criterion": "strict_improvement",  # | "improvement_or_equal"
-        "cache_evaluation": False,       # opt-in: cache identical (candidate, example) evals
+        "cache_evaluation": False,       # opt-in: cache (candidate, split, example) evals.
+                                         # Distinct valset is isolated from trainset ids.
+                                         # valset=None still shares minibatch rollouts.
         "capture_stdio": False,          # opt-in: route evaluator print() output into feedback
         "raise_on_exception": True,      # False → evaluator exceptions become score 0 + info["error"]
         # "write_agent_state": True,     # agent-readable iterations/ + pareto/ tree under run_dir
--- a/.claude/skills/gepa-optimize-anything/references/gotchas.md
+++ b/.claude/skills/gepa-optimize-anything/references/gotchas.md
@@ -61,6 +61,9 @@ work:
   emitting cache-hitting candidates without consuming eval budget and can spin until your process
   times out, still spending proposer-LLM tokens. With caching on, `stop_at_score` and/or
   `max_token_cost` (plus a wall-clock `timeout` on the launched process) are mandatory.
+  A distinct `valset` is a separate cache namespace from the trainset (list-position ids would
+  otherwise collide). `valset=None` still shares cached rollouts with minibatches. Resuming a
+  `run_dir` whose cache predates that namespacing drops the old entries.
 Agentic backends also spend LLM tokens between evals — cap them with `max_token_cost`
 (enforced as `--max-budget-usd`).
 
--- a/src/gepa/api.py
+++ b/src/gepa/api.py
@@ -190,7 +190,14 @@ def optimize(
     - write_agent_state: When True, write an agent-readable directory tree under `run_dir` alongside `gepa_state.bin` (`iterations/<id>/` + `pareto/`). Each loop iteration gets its own subdir (accepted or rejected) with `meta.json`, `components/`, `trace.json` (before/after scores + trajectories); accepted ones also get `val_scores.json`, `outputs/`, `trajectories/`. Seed is `iterations/seed/`. Default False; turn on when an agent (e.g. Claude Code) will read the run directory.
 
     # Evaluation caching
-    - cache_evaluation: Whether to cache the (score, output, objective_scores) of (candidate, example) pairs. If True and a cache entry exists, GEPA will skip the fitness evaluation and use the cached results. This helps avoid redundant evaluations and saves metric calls. Defaults to False.
+    - cache_evaluation: Whether to cache the (score, output, objective_scores) of
+      (candidate, split, example) triples. If True and a cache entry exists, GEPA skips the
+      fitness evaluation. Trainset minibatches and valset evaluation share the cache only when
+      they use the same DataLoader instance (valset=None, or the same dataset object passed as
+      both trainset and valset). A distinct valset is a separate namespace, so held-out scores
+      cannot be served from trainset rollouts at the same list index. Defaults to False.
+      Resuming a run_dir whose cache was written before this namespacing existed drops those
+      entries; already-recorded valset scores in the pickled state are left as they are.
 
     # Reproducibility
     - seed: The seed to use for the random number generator.
--- a/src/gepa/api.py
+++ b/src/gepa/api.py
@@ -18,7 +18,7 @@ from gepa.core.adapter import DataInst, GEPAAdapter, ProposalFn, RolloutOutput, 
 from gepa.core.data_loader import DataId, DataLoader, ensure_loader
 from gepa.core.engine import GEPAEngine
 from gepa.core.result import GEPAResult
-from gepa.core.state import TRAINSET_CACHE_SPLIT, VALSET_CACHE_SPLIT, EvaluationCache, FrontierType
+from gepa.core.state import EvaluationCache, FrontierType
 from gepa.logging.experiment_tracker import create_experiment_tracker
 from gepa.logging.logger import Logger, LoggerProtocol, StdOutLogger
 from gepa.proposer.merge import MergeProposer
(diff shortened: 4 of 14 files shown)


Benzhang2004 (commented):
Thanks for catching this! I pushed a small follow-up commit that moves the split derivation into the engine and makes split a required keyword argument on the cache entry points. Let me know if those make sense.

LakshyAAAgrawal:
@greptileai

LakshyAAAgrawal:
@greptileai

LakshyAAAgrawal:
Thank you very much @blerchin for identifying the issue, as well as providing a fix for it!

blerchin:
> Thank you very much @blerchin for identifying the issue, as well as providing a fix for it!

Thank you @LakshyAAAgrawal for the followup and for getting it ready to merge!

PR #430, opened 2026-08-14
Fix trace-aware batch evaluation for accepted candidates

### LINKED ISSUE #428: Full-val candidate evaluation incorrectly forces capture_traces=True
GEPA 0.1.4 incorrectly enables trace capture for an accepted candidate's full-validation evaluation.

Standalone no-network/no-LM repro:

```python
from importlib.metadata import version

import gepa
from gepa.core.adapter import EvaluationBatch


class Adapter:
    def __init__(self):
        self.calls = []
        self.propose_new_texts = self.propose

    def evaluate(self, batch, candidate, capture_traces=False):
        self.calls.append(
            (batch[0]["split"], candidate["prompt"], capture_traces)
        )
        score = 0.1 * len(candidate["prompt"])
        return EvaluationBatch(
            outputs=["ok"] * len(batch),
            scores=[score] * len(batch),
            trajectories=[{}] * len(batch) if capture_traces else None,
        )

    def make_reflective_dataset(self, candidate, evaluation, components):
        return {component: [{}] for component in components}

    def propose(self, candidate, dataset, components):
        return {"prompt": candidate["prompt"] + "x"}


adapter = Adapter()
gepa.optimize(
    seed_candidate={"prompt": "a"},
    trainset=[{"split": "train"}],
    valset=[{"split": "val"}],
    adapter=adapter,
    max_metric_calls=4,
    reflection_lm=None,
)

print("gepa", version("gepa"))
for call in adapter.calls:
    print(call)
```

Actual output (excluding progress logs):

```text
gepa 0.1.4
('val', 'a', False)
('train', 'a', True)
('train', 'ax', True)
('val', 'ax', True)
```

The expected final line is `('val', 'ax', False)`. Seed full-val is untraced and reflective minibatches are correctly traced, but accepted-candidate full-val is incorrectly traced through `default_batch_evaluate()`.

This affects DSPy because traced mode selects `bootstrap_trace_data(raise_on_error=False)` rather than `Evaluate`. A runtime-failing validation example can disappear, after which GEPA indexes missing outputs/scores and raises `IndexError`.

A narrow fix would keep reflective minibatches traced/batched while restoring accepted-candidate full-val evaluation through `adapter.evaluate(..., capture_traces=False)`.

isaacbmiller:
## Summary

- make batch evaluation trace-aware by forwarding `capture_traces` through the adapter contract and Optimize Anything adapter
- preserve batching for seed and accepted-candidate full-validation evaluation while passing `capture_traces=False`
- require the explicit `batch_evaluate(items, *, capture_traces=True)` contract and remove runtime signature inspection
- stop suppressed-exception retries when an iteration makes no metric-budget progress, with `ErrorEvent.will_continue` matching the actual outcome
- document the zero-progress `raise_on_exception=False` behavior in the public API
- document and test that direct `OptimizeAnythingAdapter.evaluate()` calls return trajectories only when trace capture is requested
- add regression coverage for issue #428, trace-aware routing, contract errors, error propagation, and stalled retries

## Batch adapter contract

The batch contract is `batch_evaluate(items, *, capture_traces: bool = True)`: seed and accepted-candidate full-val evaluation pass `False`, while reflection passes `True`.

Adapters implementing the older one-argument `batch_evaluate(items)` contract must add the keyword parameter; otherwise Python raises a clear `unexpected keyword argument 'capture_traces'` error. Wrappers and mocks must expose and forward the same contract. Adapters implementing only `evaluate()` remain supported because the sequential fallback controls trace capture. Built-in adapters are updated in this PR. Adapter exceptions propagate without retry.

Direct `evaluate()` calls also default `capture_traces` to `False`: scores and outputs are returned normally, while trajectories require explicit trace capture. With `write_agent_state=True`, full-val evaluation intentionally remains traced and sequential.

When `raise_on_exception=False`, GEPA continues only if the failed iteration consumed additional metric budget. A zero-progress failure is re-raised instead of looping indefinitely, and `ErrorEvent.will_continue` reports that decision accurately. This caveat is documented on the public `optimize()` API.

This supersedes the closed rollback PR #429.

Fixes #428

## Local verification

- Lakshya's delegating-wrapper reproduction immediately raises the expected contract `TypeError`
- focused trace/batch/engine/callback tests: 150 passed
- final docs simplification checks: 40 passed
- `uv run --frozen pytest` (675 passed, 4 skipped)
- `uv run --frozen ruff check src tests`
- `uv run --frozen pyright src` (0 errors, 0 warnings)
- `uv run --with-requirements requirements.txt mkdocs build` from `docs/`
- `git diff --check`

### DIFF
--- a/docs/docs/guides/parallel-proposals.md
+++ b/docs/docs/guides/parallel-proposals.md
@@ -107,14 +107,12 @@ class MyAdapter:
         ...  # e.g. fan out across threads, processes, or an eval service
 ```
 
-Reflective minibatches pass `capture_traces=True`; seed and accepted-candidate
-full-val batches pass `False`. Existing batch adapters must accept this keyword;
-adapters implementing only `evaluate()` continue using the sequential fallback.
-
-With `write_agent_state=True`, full-val evaluation intentionally remains traced
-and sequential. The reflection LM side is batched analogously: the built-in
-stateless reflector batches all per-component reflection calls via
-`litellm.batch_completion`.
+Third-party adapters that only implement `evaluate()` keep working unmodified
+(`default_batch_evaluate` is the sequential fallback). Batch adapters receive
+`capture_traces=True` for reflection and `False` for seed and accepted-candidate
+full-val evaluation, so existing implementations must accept the new keyword.
+The reflection LM side is batched analogously: the built-in stateless reflector
+batches all per-component reflection calls via `litellm.batch_completion`.
 
 ### `batch_evaluator`: one external call for the whole batch
 
--- a/docs/docs/guides/parallel-proposals.md
+++ b/docs/docs/guides/parallel-proposals.md
@@ -112,6 +112,11 @@ full-val batches pass `False`. Adapters implementing the older one-argument
 `batch_evaluate(items)` contract must add the keyword argument. Adapter
 exceptions propagate without retry.
 
+The same rule applies to direct `evaluate()` calls, whose `capture_traces`
+default is `False`: scores and outputs are always returned, while trajectories
+are returned only when explicitly requested. `OptimizeAnythingAdapter` now
+follows this contract instead of always returning its `side_info` as trajectories.
+
 With `write_agent_state=True`, full-val evaluation intentionally remains traced
 and sequential. The reflection LM side is batched analogously: the built-in
 stateless reflector batches all per-component reflection calls via
--- a/docs/docs/guides/parallel-proposals.md
+++ b/docs/docs/guides/parallel-proposals.md
@@ -108,9 +108,9 @@ class MyAdapter:
 ```
 
 Reflective parent/child minibatches pass `capture_traces=True`; accepted-candidate
-full-val batches pass `False`. Legacy one-argument `batch_evaluate` adapters keep
-reflective batching but use the ordered, sequential, trace-free fallback for
-full-val evaluation. Adapter exceptions propagate without retry.
+full-val batches pass `False`. Adapters implementing the older one-argument
+`batch_evaluate(items)` contract must add the keyword argument. Adapter
+exceptions propagate without retry.
 
 With `write_agent_state=True`, full-val evaluation intentionally remains traced
 and sequential. The reflection LM side is batched analogously: the built-in
--- a/docs/docs/guides/parallel-proposals.md
+++ b/docs/docs/guides/parallel-proposals.md
@@ -107,15 +107,9 @@ class MyAdapter:
         ...  # e.g. fan out across threads, processes, or an eval service
 ```
 
-Reflective parent/child minibatches pass `capture_traces=True`; accepted-candidate
-full-val batches pass `False`. Adapters implementing the older one-argument
-`batch_evaluate(items)` contract must add the keyword argument. Adapter
-exceptions propagate without retry.
-
-The same rule applies to direct `evaluate()` calls, whose `capture_traces`
-default is `False`: scores and outputs are always returned, while trajectories
-are returned only when explicitly requested. `OptimizeAnythingAdapter` now
-follows this contract instead of always returning its `side_info` as trajectories.
+Reflective minibatches pass `capture_traces=True`; seed and accepted-candidate
+full-val batches pass `False`. Existing batch adapters must accept this keyword;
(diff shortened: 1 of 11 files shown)


LakshyAAAgrawal:
@greptileai

LakshyAAAgrawal:
the shim misclassifies delegating wrappers and hard-fails them, including on the traced reflection path.

supports_trace_mode counts VAR_KEYWORD as evidence of support, so any wrapper that forwards *args/**kwargs to a legacy adapter is treated as trace-aware and GEPA passes capture_traces= straight into the inner adapter that never advertised it.

```
> import gepa
> from gepa.core.adapter import EvaluationBatch
>
> class LegacyAdapter:  # written against the shipped 0.1.4 contract
>     def __init__(self):
>         self.propose_new_texts = lambda c, r, comps: {k: c[k] + "x" for k in comps}
>     def evaluate(self, batch, candidate, capture_traces=False):
>         s = min(1.0, len(candidate["p"]) / 20.0)
>         return EvaluationBatch(outputs=[s]*len(batch), scores=[s]*len(batch),
>                                trajectories=[{}]*len(batch) if capture_traces else None)
>     def batch_evaluate(self, items):
>         return [self.evaluate(b, c, capture_traces=True) for c, b in items]
>     def make_reflective_dataset(self, c, e, comps):
>         return {k: [{}] for k in comps}
>
> class InstrumentedAdapter:  # a logging/metrics wrapper that delegates
>     def __init__(self, inner):
>         self._inner = inner
>         self.propose_new_texts = inner.propose_new_texts
>     def evaluate(self, *a, **k): return self._inner.evaluate(*a, **k)
>     def batch_evaluate(self, *a, **k): return self._inner.batch_evaluate(*a, **k)
>     def make_reflective_dataset(self, *a, **k): return self._inner.make_reflective_dataset(*a, **k)
>
> gepa.optimize(seed_candidate={"p": "s"},
>               trainset=[{"i": i} for i in range(4)], valset=[{"i": i} for i in range(4)],
>               adapter=InstrumentedAdapter(LegacyAdapter()),
>               max_metric_calls=60, reflection_lm=None, display_progress_bar=False)
> 
```

```
> main  -> OK (7 candidates)
> #430  -> TypeError: LegacyAdapter.batch_evaluate() got an unexpected keyword argument 'capture_traces'
> 
```

It fires on the capture_traces=True reflection path, not just full-val. With raise_on_exception=False it doesn't crash, it spins: the same TypeError is logged 880,411 times and the run never terminates, because no metric calls are ever charged so the budget never advances.

Legacy adapters lose full-val parallelism silently. Measured 0.99s → 3.63s (3.7×) on a 16-example valset at 20 ms/example; it scales with valset × accepted candidates × latency. It's documented in prose but nothing says anything at runtime — a one-time warnings.warn naming the adapter class would make the tradeoff visible. (Or parallelize the fallback across items, which recovers most of it without giving up trace correctness.)

OptimizeAnythingAdapter.evaluate() now returns trajectories=None by default, since capture_traces defaults to False and _assemble_no_refiner_batch honors it. That's arguably a correctness fix, but it is user-visible for anyone calling evaluate() directly — tests/test_refiner.py needed capture_traces=True added for exactly this reason.

Pre-existing, not yours: the seed full-val eval still goes through engine.py:164's per-program adapter.evaluate() and never reaches invoke_batch_evaluate, so a trace-aware adapter's batching is skipped for what is often a run's single most expensive step. Might be worth folding in while you're also touching this.

On the open question of keeping or drop the shim? I'd drop it, and finding 1 is the argument: a shim that silently misclassifies wrapper adapters and silently degrades legacy ones is worse than a loud TypeError telling the author exactly which keyword to add, especially given how short-lived the one-argument contract was.

isaacbmiller:
Man this fix is not as clean as I would like. After talking with the agents a good bit i do think that this is the correct solution and it appears to be at the correct layer, but still feels smelly.

I don't have much context on the larger GEPA codebase, so if there is some other path that can fix this issue, would be happy to rework around that, otherwise this seems to be a nice enough bandaid soln.

Removed the shim and fixed the other bugs you suggested.

Ready for re-review @LakshyAAAgrawal

LakshyAAAgrawal:
@greptileai

LakshyAAAgrawal:
Thank you very much @isaacbmiller for the bug fix!

PR #431, opened 2026-08-17
Fix broken star history chart

Dessalines39394:
The star history chart on the README is currently broken due to GitHub stargazer API restrictions. This switches the chart to the dera.page mirror, which uses a different data source that requires no API token, so the chart works again.


LakshyAAAgrawal:
Thanks for the PR, @Dessalines39394. We're going to decline this. We don't want the README to load images from a third-party host we don't already use.

PR #433, opened 2026-08-19
fix(windows): type-check platform-specific signal handling

deepanshu-iitm:
## Summary

Fix Windows compatibility in GEPA’s process-timeout handling.

GEPA uses Unix signal APIs to enforce in-process execution timeouts and to stop an unresponsive Claude Code subprocess. Windows does not provide several of these APIs:

- `SIGALRM`
- `setitimer`
- `ITIMER_REAL`
- `alarm`
- `SIGKILL`

Direct references to them caused `uv run pyright src` to fail on Windows. More importantly, when AutoResearch tried to stop an unresponsive subprocess, the cleanup path could fail before it reached the normal `proc.kill()` fallback.

## What changed

- Resolve `SIGALRM` before using the in-process signal timeout mechanism, so the Unix-only code remains guarded on Windows.
- Keep the Pyright platform-specific suppressions limited to the guarded Unix signal calls.
- Use a non-`SIGTERM` fallback value when `SIGKILL` is unavailable. This preserves the existing behavior: on Windows, `_signal_claude_process()` uses `proc.kill()` for the forced-stop step.
- Add a regression test that simulates:
  1. Windows platform behavior,
  2. an unavailable `SIGKILL`,
  3. a subprocess that ignores `terminate()`.

  The test verifies that GEPA then calls `kill()`.

## Why this matters

Windows contributors can now run GEPA’s required Pyright check successfully. AutoResearch’s watchdog also reliably terminates an unresponsive Claude Code process on Windows instead of failing while looking up a Unix-only signal constant.

## Validation

```text
uv run ruff check src/gepa/utils/code_execution.py src/gepa/oa/engines/autoresearch.py tests/test_optimize_anything_autoresearch_engine.py
uv run pyright src
uv run pytest tests/test_optimize_anything_autoresearch_engine.py -q

### DIFF
--- a/src/gepa/oa/engines/autoresearch.py
+++ b/src/gepa/oa/engines/autoresearch.py
@@ -165,7 +165,9 @@ def _terminate_claude_process(running: _RunningClaudeProcess) -> tuple[str, str]
         try:
             proc.wait(timeout=5)
         except subprocess.TimeoutExpired:
-            _signal_claude_process(proc, signal.SIGKILL)
+            # SIGKILL is not defined by Windows' signal module. Any value other
+            # than SIGTERM uses the direct `proc.kill()` fallback there.
+            _signal_claude_process(proc, getattr(signal, "SIGKILL", -9))
             proc.wait()
     return _collect_claude_output(running)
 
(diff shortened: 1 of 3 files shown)


LakshyAAAgrawal:
Dear @deepanshu-iitm, thank you so much for identifying and fixing the bug!

PR #435, opened 2026-08-21
fix(reflection): skip a truncated reflection output instead of adopting it as a candidate

### LINKED ISSUE #390: Truncated reflection LM output (missing code fence) is silently adopted as a candidate prompt — should format failures be handled on the reflection side as in #265?
**1. Background**

I have been reproducing the Qwen3-8B AIME experiment from the GEPA paper using the released artifact configuration (8,192-token server context; thinking mode, which the sampling parameters temperature=0.6 / top_p=0.95 / top_k=20 suggest). The reproduction itself was successful — I obtained a baseline of 24.0% and a post-GEPA test score of 31.33% (AIME-2025, 30 problems × 5 runs, seed 0), close to the paper's reported 27.33 / 32.00. While running it, however, I observed an unexpected behavior on the reflection LM side that I would like to share and ask about.

**2. What I observed**

For AIME, the reflection input (execution traces plus solution feedback) alone is large: measuring with the Qwen3 tokenizer, the median was about 5,900 tokens, which leaves only ~2K tokens of the 8,192 budget for generation. As a result, the reflection LM frequently ran out of budget mid-thinking, and its output was cut off mid-sentence without a closing `</think>` or a closing code fence. These truncated thinking monologues were then adopted verbatim as new candidate prompts. In my run, 13 out of 23 candidates were incomplete monologues starting with `<think>`, and notably, the final best candidate was one of them — a ~2,300-character text beginning with "Okay, let me try to figure out how to improve the parameter value based on the evaluation data provided..." that cuts off mid-sentence. Interestingly, this candidate scored 40.0% on val and 31.33% on test, the closest of all candidates to the paper's reported 32.00.

**3. Where this comes from in the code**

Tracing the cause, `output_extractor` in `src/gepa/strategies/instruction_proposal.py` returns the raw LM output when no fence pair is present:

````python
def extract_instruction_text() -> str:
    # Find the first and last backtick positions (if any)
    start = lm_out.find("```") + 3
    end = lm_out.rfind("```")

    # Handle if the first and last backticks are the same or overlap
    if start >= end:
        # Handle incomplete blocks
        stripped = lm_out.strip()
        ...
        return stripped   # <- truncated output with no fence becomes new_instruction verbatim
````

When the output completes normally, only the fenced instruction is extracted, so there is no issue. But when a thinking model's output is truncated at the context limit, the fences are never generated, this fallback path is taken, and the incomplete `<think>` monologue becomes the instruction wholesale. From there it only needs to pass the minibatch score gate to stay in the candidate pool, so everything proceeds silently with no error or warning.

**4. Suggestion — applying the principle of #265 to the reflection side**

This seems closely related to #265, which proposes treating format failures of the task LM (e.g., malformed JSON) as hard failures — score 0 with an explicit parse-error feedback — rather than letting them pass silently. The current situation on the reflection side is exactly the asymmetric case: a format failure of the reflection LM (missing fence due to truncation) is not detected at all, and the malformed output is adopted as a candidate as-is. Following the same principle, when the extractor finds no fence pair (or detects an unterminated `<think>` block), the proposal could be treated as failed — skipping the iteration with a logged warning, or retrying — instead of adopting the raw text. This would prevent truncated reasoning monologues from silently entering the candidate pool.

**5. Questions**

First, is the current fallback — using the raw output as the instruction when no fence is present — intended behavior, or was it written under the assumption that a well-formed output always contains a fence? Second, in the original Qwen3-8B AIME run, were there cases where truncated `<think>` text was adopted as a candidate, and could the AIME results reported in the paper have come from candidates of this form? The artifact includes the gpt-4.1-mini AIME logs and the Qwen logs for other tasks, but the Qwen AIME logs are not included, so I was not able to check this directly on my side. Third, would you consider extending the format-validation principle of #265 to the reflection LM's output as well?

**6. Environment note**

My reproduction used plain vLLM (max-model-len 8192) and dspy 2.6.23, rather than Arbor and the vendored dspy from the artifact, so some of the observed frequencies may differ from the original setup. The code path described in section 3, however, is in the current `src/gepa` library and is independent of that environment. If the original experiment handled this differently, I would appreciate any pointers.

overgoy:
Resolves #390.

Handles truncated reflection outputs so unfinished reasoning is no longer passed through as the next instruction. Truncated proposals are now skipped safely, while existing extraction behavior stays unchanged.

Added coverage for truncated, well-formed, and salvage cases, including reflect, reflect_many, and the real proposer flow.

I used Claude Code to make this change: it edited the repository directly and the project's own test suite was run before I sent it. I have reviewed it and can explain it.

### DIFF
--- a/src/gepa/proposer/reflective_mutation/reflection_lm.py
+++ b/src/gepa/proposer/reflective_mutation/reflection_lm.py
@@ -183,6 +183,18 @@ class StatelessReflectionLM:
 
         proposals = [ReflectionProposal(new_texts={}, prompts={}, raw_lm_outputs={}) for _ in jobs]
         for (job_idx, name, prompt, _messages), raw_output in zip(rendered, raw_outputs, strict=True):
+            # A reflection that ran out of generation budget mid-thought has no
+            # instruction to extract, and the extractor would hand its raw
+            # monologue back as the proposal. Leave the component out instead:
+            # the proposer already treats a missing component as "no update"
+            # and keeps the parent text, so nothing unreviewed enters the pool.
+            if InstructionProposalSignature.is_truncated_output(raw_output):
+                self._log(
+                    f"Component '{name}': the reflection output ends inside an unterminated reasoning "
+                    "block, so the generation was truncated and carries no proposal. Skipping. Raise the "
+                    "reflection LM's output budget, or shrink the reflection minibatch, to recover it."
+                )
+                continue
             new_instruction = InstructionProposalSignature.output_extractor(raw_output.strip())["new_instruction"]
             proposals[job_idx].new_texts[name] = new_instruction
             proposals[job_idx].prompts[name] = prompt
--- a/src/gepa/strategies/instruction_proposal.py
+++ b/src/gepa/strategies/instruction_proposal.py
@@ -45,21 +45,33 @@ Provide the new instructions within ``` blocks."""
     def is_truncated_output(cls, lm_out: str) -> bool:
         """Whether ``lm_out`` was cut off before the model finished reasoning.
 
-        A reflection LM that hits its generation budget mid-thought emits an
-        opening reasoning tag it never closes and never reaches the fenced
+        A reflection LM that hits its generation budget mid-thought opens a
+        reasoning block, never closes it, and never reaches the fenced
         instruction, so its monologue arrives at :meth:`output_extractor` with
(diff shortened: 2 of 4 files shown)


LakshyAAAgrawal:
Thank you @overgoy for your PR! Your skip-on-truncated approach for #390 the final fix we landed in #462 (along with some changes proposed in #460). Closing as superseded by #462 (credited in #465)!

PR #444, opened 2026-08-29
Add gepa.rpc: bidirectional gRPC server + TS/Rust SDKs for non-Python evaluators

anikag35:
Adds `gepa.rpc`, a gRPC server in front of `gepa.optimize` and `optimize_anything`, so TypeScript and Rust developers can drive GEPA optimization while writing their evaluator in their own language, with no Python required client-side. It exposes two RPCs: `RunOptimization`, which wraps `gepa.optimize`, and `RunOptimizationOmni`, which wraps `optimize_anything` and is selectable across all four registered engines (`gepa`, `autoresearch`, `best_of_n`, `meta_harness`) via a new `engine` field. TypeScript (`@gepa/sdk`) and Rust (`gepa-sdk`) client SDKs are included, both fully implementing both RPCs, along with a new `gepa[rpc]` extra, a `gepa-rpc` console script, and a Dockerfile for standalone deployment. `OmniEvaluateBatchRequest.opt_states` carries each example's warm-start history (top-K best prior evaluations) to the client, matching the `opt_state` pattern already used by `examples/blackbox` and `examples/circle_packing`.

## Test plan

- [x] Full `pytest`, `ruff`, `pyright` clean against this repo's own config (719 existing tests + 25 new, all passing)
- [x] Real GitHub Actions run of `run_tests.yml` against this branch — 3.10–3.14 matrix all green
- [x] Both SDKs build and run live against a real server (not mocked)
- [x] All four Omni engines run live end-to-end with real LLM calls, including the `autoresearch`/`meta_harness` agent subprocesses
- [x] Reflective mutation, seedless mode (LLM-proposed seed), `opt_states`, `valset`, disconnect/resume via `run_dir` checkpointing, graceful shutdown (SIGTERM), and error paths (Python/TS/Rust) all verified live
- [x] `docker build` and `docker run` verified via GitHub Actions (image starts, binds, and listens)
- [x] Found and fixed 2 real bugs only CI caught: a stale Rust toolchain pin (`1.75` → `stable`, blocked by a transitive `edition2024` requirement) and 2 pre-existing Rust closure type-inference errors


anikag35:
Pushed a local record/replay stand-in for the LLM provider (gepa.rpc.testing.fake_llm_server), keyed by (model, messages) the same way this repo's own mocked_lms fixture is, so litellm can be pointed at it via OPENAI_API_BASE instead of hitting a live provider; and one e2e test per language (tests/test_rpc_e2e/, sdk/typescript/tests/e2e.ts, sdk/rust/tests/e2e.rs), each spawning a real gepa-rpc subprocess and driving it through that language's real client.

There is no mocking of gepa.optimize/optimize_anything anywhere in these three tests, unlike the existing test_rpc_integration.py.

Verified locally: all three pass against a single shared, committed cache and golden file, and TS/Rust reproduce Python's exact recorded prompts (a cache hit, not just a matching final output), which is direct evidence their wire encoding and the server's optimizer state transitions are identical. Also did a second, independent live recording triggered entirely from the Rust client against a real API key, to confirm it drives a genuinely fresh optimization rather than only replaying Python's result.

Test plan for this push:
- [x] All three e2e tests pass locally (Python via pytest, TypeScript via npm run test:e2e, Rust via cargo test --test e2e)
- [x]  Existing test_rpc_integration.py (25 tests) still passes unmodified
- [x] Cross-language consistency confirmed via cache-hit determinism, not just matching final output. 
- [x] Rust path additionally verified with its own independent live recording against a real API key
- [ ] Real GitHub Actions run

PR #445, opened 2026-09-01
fix: store the candidate a refiner actually evaluated, not the pre-re…

### LINKED ISSUE #440: Refiner score is credited to the unrefined candidate
## Bug

When `RefinerConfig` is enabled, `OptimizeAnythingAdapter` can return the score of a refined candidate, but `ReflectiveMutationProposer` continues to forward the original unrefined candidate. The candidate pool can therefore store a candidate with a validation score it cannot reproduce without another refinement.

## Environment

- Confirmed with GEPA 0.1.1, Python 3.14.6
- The same control flow appears present on current `main` as of 2026-08-26

## Minimal reproduction

```python
import json
from gepa.adapters.optimize_anything_adapter.optimize_anything_adapter import OptimizeAnythingAdapter
from gepa.optimize_anything import RefinerConfig

refinements = iter(["FIX1", "FIX2", "FIX3"])
refiner = lambda prompt: json.dumps({"code": next(refinements)})

def evaluator(candidate, **kwargs):
    scores = {"BROKEN": 0.0, "FIX1": 1.0, "FIX2": 2.0, "FIX3": 3.0}
    code = candidate["code"]
    side_info = {"error": "cannot run"} if code == "BROKEN" else {"ran": code}
    return scores[code], None, side_info

adapter = OptimizeAnythingAdapter(
    evaluator=evaluator,
    parallel=False,
    refiner_config=RefinerConfig(refiner_lm=refiner, max_refinements=3),
    cache_mode="off",
)
original = {"code": "BROKEN", "refiner_prompt": "repair errors"}
batch = adapter.evaluate([None], original)
print(batch.scores[0])        # 3.0
print(batch.outputs[0][1])   # FIX3 candidate
```

At the adapter boundary the result is correct: score `3.0` and the winner is `FIX3`. However, the outer proposer constructs `CandidateProposal(candidate=new_candidate)` and only copies `child_eval.scores`/`child_eval.outputs` into the evaluation metadata. The engine then full-evaluates and stores `proposal.candidate`, which is still `BROKEN`.

## Expected behavior

If a refined candidate wins, the winning candidate object should be the candidate passed to full validation and stored in the candidate pool, so the stored candidate and recorded score refer to the same artifact.

## Additional impact

A zero-scoring unrunnable proposal can be accepted because its refined attempt beats the parent, while the unrunnable proposal itself is what enters the pool. Later evaluations may refine it again to a different candidate, making scores non-reproducible and lineage misleading.

dhairya2006-del:
Fixes #440

## Problem
When `RefinerConfig` is enabled, `OptimizeAnythingAdapter` could return a
refined candidate's score while `ReflectiveMutationProposer` still stored
the pre-refinement candidate. The candidate pool ended up holding a
candidate paired with a score it couldn't reproduce.

## Fix
Per the direction confirmed in #440 (option 1):

- Added `EvaluationBatch.evaluated_candidates` (optional, defaults to
  `None`) - adapters can populate this when the candidate they actually
  scored differs from the one they were given. Every adapter that doesn't
  do internal refinement is unaffected.
- `OptimizeAnythingAdapter.evaluate()` now populates it with the winning
  refined candidate when `refiner_config` is set.
- `ReflectiveMutationProposer` now prefers `evaluated_candidates[idx]`
  over the pre-refinement candidate when building each `CandidateProposal`,
  falling back to the original behavior when it's `None`.

## Testing
- Added `test_evaluated_candidates_matches_refined_score` in
  `tests/test_refiner.py`, adapted from the issue's minimal repro —
  asserts the stored candidate reproduces the recorded score after
  refinement.
- `uv run pytest tests/` - all tests pass except 13 pre-existing,
  unrelated Windows-only failures (mlflow URI handling, console unicode
  encoding) present before this change.
- `uv run pyright` and `uv run pre-commit run` both clean on all changed
  files.

### DIFF
--- a/src/gepa/adapters/optimize_anything_adapter/optimize_anything_adapter.py
+++ b/src/gepa/adapters/optimize_anything_adapter/optimize_anything_adapter.py
@@ -154,8 +154,7 @@ class BatchEvaluatorWrapper:
                     # Mirror the evaluator path's loudness instead of silently
                     # discarding user diagnostics.
                     raise TypeError(
-                        f"batch_evaluator result {idx}: side_info must be a dict or None, "
-                        f"got {type(si_raw).__name__}"
+                        f"batch_evaluator result {idx}: side_info must be a dict or None, got {type(si_raw).__name__}"
                     )
                 # Defensive copy (mirror of EvaluatorWrapper): the user may
                 # retain and mutate their dict; the cache, best-evals history,
@@ -397,6 +396,10 @@ class OptimizeAnythingAdapter(GEPAAdapter):
         side_infos: list[SideInfo] = [info for _, _, info in eval_output]
         # outputs = list of (score, best_candidate, side_info) tuples
         outputs = [out for _, out, _ in eval_output]
+        # The refiner may have swapped in a different candidate than the one
+        # passed in; surface it so callers can store the candidate that
+        # actually produced this score, instead of assuming it's `candidate`.
+        evaluated_candidates = [best_candidate for _, best_candidate, _ in outputs]
         objective_scores = [self._extract_objective_scores(candidate, si) for si in side_infos]
 
         # Count actual evaluator invocations from the attempt history recorded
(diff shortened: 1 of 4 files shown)


youph:
Is this still being reviewed?

LakshyAAAgrawal:
Hi @dhairya2006-del, would you be able to see the changes I made in #457 and if it aligns with your PR? I can next go ahead and merge both.

Issue #448, opened 2026-09-03
Bugs Identified Post 0.1.4


LakshyAAAgrawal:
# Draft: upstream issue for gepa-ai/gepa (NOT FILED)

Status: draft written 2026-09-03 against `gepa/main` @ `0632cdb5` (2026-09-01), compared
with the 0.1.4 release merge `665cbc36`. Every observation below was reproduced with the
script at the end, run once against a fresh install of `main` and once against this fork
(0.1.4 + ported fixes #425/#430). File only after review; keep the repro verbatim.

---

## Title

Legacy `optimize_anything(config=GEPAConfig)` path after #346: budget now aborts
mid-iteration and discards paid evaluations; `_budget` leaks into reflection prompts; every
run starts an HTTP listener and writes an `outputs/` tree under CWD

## Summary

Since #346 (`6d897c1e`, "engine-pluggable optimize_anything") a call that still uses the
0.1.x API

```python
from gepa.optimize_anything import optimize_anything, GEPAConfig, EngineConfig
optimize_anything(seed_candidate=..., evaluator=..., dataset=..., valset=...,
                  config=GEPAConfig(engine=EngineConfig(max_metric_calls=N, run_dir=...)))
```

is converted by `_from_legacy_config` (`src/gepa/optimize_anything.py:274-300`) and run
through `EvalServer` + `BudgetTracker` (`optimize_anything.py:236-246`). The conversion
keeps the config field-for-field, but the runtime around GEPA core changed in ways that
alter results, prompts and side effects for unchanged user code. None of this is mentioned
in the #346 description or the docs for `GEPAConfig`.

## Observed (same script, same seed; details of the script below)

| observation                                   | 0.1.4 (`665cbc36`)     | main (`0632cdb5`)                     |
|-----------------------------------------------|------------------------|---------------------------------------|
| cap 10: evaluator calls / candidates          | 8 / 2 (val 0.2, 1.0)   | 8 / 2 (val 0.2, 1.0)                  |
| cap 7: evaluator calls / candidates           | 8 / 2 (val 0.2, 1.0)   | 7 / **1** (val 0.2), `BudgetExhausted` mid-valset |
| cap 7: `result.total_metric_calls`            | 10                     | 4 (state from before the iteration)   |
| `"_budget"` appears in a reflection prompt    | no                     | **yes**                               |
| HTTP listener started for the run             | no                     | **yes** (`localhost:<ephemeral>`)     |
[...]

PR #452, opened 2026-09-08
fix(oa): stop the eval budget from discarding paid evals; keep _budget out of reflection prompts

Benzhang2004:
Addresses items 1 and 2 of #448.

## Problem

Since #346 the optimize_anything eval server enforced `max_evals` as a hard stop inside `BudgetTracker.record()`, which raised `BudgetExhausted` from the middle of GEPA core's minibatch and valset evaluations. The in-flight child, already scored on most of the valset, was thrown away and the result reloaded from the previous iteration's state. A legacy `GEPAConfig` run with `max_metric_calls=7` returned only the seed where 0.1.4 returned the improved child. `evaluate_batch`'s own docstring promised an iteration-boundary overshoot that `record()` did not allow, dropping results the user had already paid for.

Separately, `evaluate()` and `evaluate_batch()` injected `_budget` into every per-example `info`. The adapter copies every `side_info` key into the reflective dataset, so budget bookkeeping reached the reflection LM in every prompt.

## Changes

- `BudgetTracker.record()` never raises. `check(needed=1)` is the only gate and now also covers the whole-split pre-check in `evaluate_examples`.
- `BudgetTracker` gains `enforce`. `GepaEngine` sets it to `False` after handing `max_evals` to core's `MaxMetricCallsStopper`, restoring the launcher's iteration-boundary semantics for both the legacy `GEPAConfig` path and `engine="gepa"`. External engines keep hard enforcement.
- `evaluate()` / `evaluate_batch()` no longer inject `_budget` into per-example info. Budget status stays on `/status`, the `evaluate_examples` envelope, and the final result.
- Document the two cap semantics on `OptimizeAnythingConfig.max_evals`.

## Verification

Issue repro at cap 7, before vs after:

| observation | before | after (matches 0.1.4) |
|---|---|---|
| evaluator calls / candidates | 7 / 1 | 8 / 2 (val 0.2, 1.0) |
| `total_metric_calls` | 4 | 10 |
| `_budget` in a reflection prompt | yes | no |

Tests added:
- `tests/test_budget_tracker.py`: check/record/needed/enforce unit tests.
- `tests/test_optimize_anything_eval_server.py`: a batch crossing the cap records every pair; per-example info carries no server bookkeeping.
- `tests/test_optimize_anything_gepa_config.py`: the #448 repro as a regression test.

Full suite: 706 passed, 4 skipped. `ruff check`, `ruff format --check`, and `pyright src/` are clean.

## Not in this PR

Items 3 (lazy HTTP listener, `outputs/` under CWD) and 5 (core caching transient failures) from #448 are planned as separate PRs.


G-ChandraSekhar:
Hi @Benzhang2004 — I'd like to contribute the core transient-failure caching fix from #448. I saw your note in #452 that you planned a separate PR for it. Have you started that work, or would it be useful for me to take it on?

I reproduced the persistence problem on `main` at `15ee314f`: a `BatchEvaluatorWrapper(..., raise_on_exception=False)` converts a temporary batch exception into a marked zero score. A seed-only `gepa.optimize(..., cache_evaluation=True, max_metric_calls=1)` run writes it into `gepa_state.bin`. After loading the state, a lookup through `EvaluationCache.evaluate_with_cache_full` returns `0.0` with zero new evaluations, while a direct retry through the same adapter returns `1.0`. The reproduction uses a deterministic fake evaluator and no LLM/network calls.

My proposed scope would cover the seed, parent/child minibatch, and full-validation cache writes, with regression coverage for persisted-cache retries and legitimate zero scores remaining cacheable. One design question: would you prefer explicit optional per-example cacheability metadata on `EvaluationBatch`, rather than having generic core code interpret the optimize-anything adapter's private output marker? We'd also need to agree on handling already-contaminated checkpoints.

Happy to align with your existing plan or contribute tests to your branch if implementation is already underway.

Benzhang2004:
Hi @G-ChandraSekhar, thanks for the thorough repro. Please take it on. I have not started the caching work.

On the design question: I'd go with explicit per-example cacheability metadata on `EvaluationBatch` rather than having core read the adapter's private `_gepa_transient_failure` marker. An optional `cacheable: list[bool] | None` (default `None` meaning all cacheable) keeps core adapter-agnostic, and the optimize-anything adapter sets it to `False` for its synthetic failure results. Core's `put_batch` call sites can then skip unmasked entries. Your proposed scope of the seed pass, parent/child minibatch, and full-validation writes sounds right.

On already-contaminated checkpoints: I'd keep it simple. Don't try to detect or scrub them on load. There's no reliable signal in the persisted score alone, and a scrub that reads the marker out of cached outputs would reintroduce the coupling we're removing. A short note in the changelog that checkpoints written before the fix may hold transient-failure zeros, and that deleting `gepa_state.bin` clears them, is enough. If you want to go further, a warning at load time when a cached output carries the marker would be a reasonable middle ground, but I wouldn't block the PR on it.

Please also cover the two cases you mentioned in tests: a persisted-cache retry re-evaluates after a transient failure, and a legitimate 0.0 score stays cached. If you reference #448 in the PR description and link it back here, I'll review when it's up.

G-ChandraSekhar:
Hi @Benzhang2004, the caching fix is up in #456. It follows the agreed design: optional per-example `cacheable` metadata, adapter-local translation of the transient-failure marker, and propagation through seed, parent/child minibatch, and full-validation cache writes.

Regression coverage includes persisted-cache retries with and without tracing, legitimate zero scores remaining cached, and all four evaluation paths. The full suite passed locally with 821 tests and no skips using the optional dependency versions listed in the PR; pre-commit and type checking with the locked CI dependencies also pass. The PR documents the separate MCP compatibility findings.

I put the older-checkpoint warning in the existing resume FAQ because the repository has no changelog. It explicitly notes that restarting clears saved progress. This addresses item 5 of #448 and leaves the budget work in this PR separate. Ready for your review—thanks for the guidance.

PR #453, opened 2026-09-11
fix: skip seed valset evaluation when resuming from run_dir

dbreunig:
## The bug

When a run resumes from a `run_dir` that already holds `gepa_state.bin`, `GEPAEngine.run` evaluates the seed candidate on the full validation set and then discards the result. The seed evaluation runs before the saved state is checked, and the resume branch of `initialize_gepa_state` never reads `seed_valset_evaluation`. Every resume pays for a full valset evaluation that has no effect. With a 10 example valset and an expensive metric, that is ten wasted metric calls per resume.

## How to see it

Run an optimization with a `run_dir`, then call `gepa.optimize` again with the same `run_dir`. The log shows the adapter evaluating the seed on the full valset and then prints `Loading gepa state from run dir`. An adapter that counts `evaluate` calls records one call over the whole valset during the resume.

## The fix

- `GEPAEngine.run` checks for `gepa_state.bin` before the seed evaluation and skips the evaluation when the saved state will be loaded. The `on_optimization_start` callback still fires in the same place.
- `initialize_gepa_state` accepts `None` for `seed_valset_evaluation` on the resume path. It raises a `ValueError` if `None` is passed and no saved state exists. Callers that pass an evaluation keep working unchanged.
- On resume, the engine no longer rewrites the seed's `iterations/seed/` outputs or the seed cache entry. Both were already skipped or redundant on resume.
- Fresh runs are unchanged. The callback, seed evaluation, cache write, and agent state files happen in the same order as before.

## The test

`test_resume_skips_seed_valset_evaluation` in `tests/test_state.py` runs a small optimization with a `run_dir` using an adapter that counts `evaluate` calls and needs no LM. It then resumes from that `run_dir` with a fresh counting adapter and asserts:

- the adapter's `evaluate` was never called during the resume
- the resumed result's `total_metric_calls` equals the saved state's `total_num_evals`
- the resumed result's `num_full_val_evals` equals the saved state's `num_full_ds_evals`

The test fails on `main` with one evaluate call over 10 examples and passes with this change. The full suite passes (700 passed, 4 skipped).

https://claude.ai/code/session_016XBgSqxXL22qUyVsXB45XB

### DIFF
--- a/src/gepa/core/engine.py
+++ b/src/gepa/core/engine.py
@@ -804,15 +804,20 @@ class GEPAEngine(Generic[DataId, DataInst, Trajectory, RolloutOutput]):
             ),
         )
 
-        # Evaluate seed candidate on valset (after on_optimization_start callback).
-        # Policies may narrow the seed evaluation via the optional get_seed_eval_batch
-        # hook; the state does not exist yet, so get_eval_batch cannot be used here.
-        seed_batch_fn = getattr(self.val_evaluation_policy, "get_seed_eval_batch", None)
-        seed_val_ids = list(seed_batch_fn(valset)) if seed_batch_fn is not None else list(valset.all_ids())
-        seed_valset_evaluation = valset_evaluator(self.seed_candidate, seed_val_ids)
-
-        # Initialize state with pre-computed seed evaluation
+        # A saved state already holds the seed's valset scores, so a resumed run
+        # skips the seed evaluation instead of spending metric calls on a result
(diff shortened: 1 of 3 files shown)


LakshyAAAgrawal:
Thanks a lot @dbreunig!

dbreunig:
No problem. Wanted to clean up a comment or two before you got to it, but Claude posted without permission.

PR #456, opened 2026-09-15
fix: avoid caching transient evaluation failures

G-ChandraSekhar:
# fix: avoid caching transient evaluation failures

## Problem

When `BatchEvaluatorWrapper(..., raise_on_exception=False)` catches a temporary whole-batch failure, it returns synthetic `0.0` scores. The optimize-anything adapter already excludes those results from its own cache, but GEPA core could still cache them during seed evaluation, parent or child minibatch evaluation, or full validation. Once persisted in `gepa_state.bin`, the synthetic zero could be reused indefinitely instead of retrying the evaluator.

## Changes

- Add optional per-example `cacheable` metadata to `EvaluationBatch`; `None` preserves the existing behavior that every result is cacheable.
- Have the optimize-anything adapter mark only synthetic transient-failure results as uncacheable.
- Carry the mask through seed, parent/child minibatch, and full-validation cache writes while keeping core independent of adapter-private markers.
- Document that older checkpoints are not scrubbed automatically; restarting in a fresh run directory or deleting `gepa_state.bin` clears affected state and discards saved progress.

## Verification

- Added a persisted-checkpoint regression proving a transient failure is retried and returns the recovered score.
- Added coverage proving a legitimate `0.0` remains persisted and is served from cache.
- Added an end-to-end test covering seed, parent, child, and full-validation writes.
- Full suite on Python 3.11 with all four optional adapter dependencies installed: **821 passed, no skips** (`PYTHONPATH=src uv run --no-sync pytest -q`). Versions: DSPy 3.3.1, LangChain 1.4.0 / langchain-core 1.6.3, llm-structured-confidence 0.4.5, and MCP 1.30.0. Dependency files are unchanged.
- MCP 2.2.0 was also tested: two response-format tests fail identically on unchanged `main` and this branch (error flag and image MIME type handling). The all-green optional run uses MCP 1.x; it does not establish MCP 2.x compatibility.
- `uv run --no-sync ruff check src/` passes.
- `uv run --no-sync pyright` passes with zero errors in a separate environment installed with `uv sync --locked --extra dev --python 3.11`, matching CI's dependency setup. With optional MCP 1.30.0 installed, Pyright reports 14 existing MCP adapter errors on both unchanged `main` and this branch.
- All applicable repository pre-commit hooks passed (`uv run --no-sync pre-commit run --from-ref origin/main --to-ref HEAD`).

Addresses item 5 of #448. Follow-up to #452.

Design coordinated with Benzhang2004: https://github.com/gepa-ai/gepa/pull/452#issuecomment-5668181107.
The checkpoint migration note is in the existing resume FAQ because the repository has no changelog file.


G-ChandraSekhar:
Hi @Benzhang2004 — friendly nudge that this is ready for review whenever you get a chance. It follows the design from your comment on #452 and addresses item 5 of #448. Also, no CI seems to have run on the PR — let me know if there's anything I need to do on my end to trigger it.

PR #460, opened 2026-09-18
Reject malformed reflection fences

sethkimmel3:
When a reflection model's output exceeds it's allowed output length, it will often stop before generating the fenced backticks that allow gepa to post-process the results. In cases where this occurs, it creates malformed candidate prompts or leaks validation text into the candidate.


LakshyAAAgrawal:
Thank you @sethkimmel3. The reject malformed/truncated reflection output for #390 landed in #462 (including some changes proposed in #435). Closing as superseded by #462 (credited in #465).

